Issue #25, opened 2016-07-29
Proxy support


mad-mike:
- VSCode Version: 1.3.1
- OS Version: Windows 10
- REST Client Version: 0.6.2

Steps to Reproduce:
1. Enter a network environment where an HTTP proxy is required
2. Make the proxy settings in VSCode's settings.json ("http.proxy": "http://proxy:8080")
3. Try to make an HTTP request, e.g. 'GET http://example.org/'
4. There is an error message that says "Error: Error: connect ETIMEDOUT ..."

Given the fact that I can't find any discussions or related code in this project I'm assuming the REST Client currently does not support HTTP proxies.

Huachao:
@mad-mike , currently we really don't support this, and will implement this later. And it's very welcome that you can contribute to this code. :smile:

PR #27, opened 2016-08-01
Proxy support

mad-mike:
implementing #25

### DIFF
--- a/src/models/configurationSettings.ts
+++ b/src/models/configurationSettings.ts
@@ -1,4 +1,4 @@
-"use strict";
+/// <reference path="../../typings/vscode-typings.d.ts" />
 
 import { workspace } from 'vscode';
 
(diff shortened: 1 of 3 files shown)


src/models/configurationSettings.ts
```diff
@@ -1,4 +1,4 @@
-"use strict";
+/// <reference path="../../typings/vscode-typings.d.ts" />
```
Huachao: I have a question that why adding this line and removing the strict mode?

mad-mike: I found that the typescript compiler will introduce "use strict" anyway.
Then I added the ref to the type definitions in order to get IntelliSense for the vscode API.

Huachao: I found even without this line, I can still get IntelliSense with vscode

mad-mike: @Huachao, you are right. Explicit typings are only necessary for npm packages that don't contain them out of the box, as it seems (http://www.typescriptlang.org/docs/handbook/typings-for-npm-packages.html). I'm going to save a few lines of code from now on ;)

Huachao:
Merged, thanks

PR #82, opened 2017-03-19
Set language when opening new editor

dakaraphi:
The new feature is working great.  Thanks!
After using this for a bit, I found a couple of things that improve my work.

1.  I set the language automatically when opening a new editor.  I wanted to do this also for reusing the same editor, but I couldn't find an API to do it.
2.  Included request url and elapsed time in the response document
3.  Made the response information into comments which makes it play nice with the language theme and coloring.

Hopefully you can consider these changes.  Thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -348,7 +348,7 @@
         "rest-client.includeAdditionalInfoInResponse": {
           "type": "boolean",
           "default": false,
-          "description": "Include additional information such as request URL and response time"
+          "description": "Include additional information such as request URL and response time when preview is set to use untitled document"
         }
       }
     }
--- a/package.json
+++ b/package.json
@@ -339,6 +339,16 @@
           "type": "boolean",
           "default": false,
           "description": "Preview response in untitle document if set to true, otherwise displayed in html view"
+        },
+        "rest-client.previewResponseSetUntitledDocumentLanguageByContentType": {
+          "type": "boolean",
+          "default": false,
+          "description": "Attempt to automatically set document language based on Content-Type header when showing each response in new tab"
+        },
+        "rest-client.includeAdditionalInfoInResponse": {
+          "type": "boolean",
+          "default": false,
+          "description": "Include additional information such as request URL and response time"
         }
       }
     }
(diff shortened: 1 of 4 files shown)


Huachao:
@dakaraphi the code LGTM. However, I still want to discuss with you that do we give user the flexibility that  user can view the response in `http` language and not commented headers and request url?

dakaraphi:
@Huachao so are you simply asking to provide a user preference setting?
Something like `rest-client.previewResponseSetUntitledDocumentLanguageByContentType`
*  default false: Will work as it did without my changes
* true: will attempt to set language and make header info comments

I added 2 additional pieces of information.  The request url and response time.  Do you want that included when in `http` format or only when language can be set?

Huachao:
@dakaraphi that's what I mean, what do you think about it? And thanks for your PR

dakaraphi:
@Huachao I added 2 new preferences.  Let me know what you think.
For now, I have them both set as default false.

I do think setting the language adds a lot of power and makes the experience richer.
1.  I can now use my theme created for custom coloring in response
2.  I can now use all language feature extensions like JSON query or XML Xpath directly on the response.  This is pretty cool.

If vscode resolves https://github.com/Microsoft/vscode/issues/1800 it will make it even better :-)

src/views/responseUntitledFileContentProvider.ts
```diff
@@ -47,6 +95,27 @@ export class UntitledFileContentProvider {
...
+
+    private static commentBegin(language: string) {
+        const REST_RESPONSE = 'REST Response Information:';
+        let commentStyle = {
+            xml: '<!-- ',
+            json: '/* ',
+            html: '<!-- ',
+            css: '/* '
+        }
+        return commentStyle[language] ? commentStyle[language] + REST_RESPONSE: '';
```
Huachao: add a space before `:`

package.json
```diff
@@ -338,7 +338,17 @@
...
+        },
+        "rest-client.previewResponseSetUntitledDocumentLanguageByContentType": {
+          "type": "boolean",
+          "default": false,
+          "description": "Attempt to automatically set document language based on Content-Type header when showing each response in new tab"
+        },
+        "rest-client.includeAdditionalInfoInResponse": {
+          "type": "boolean",
+          "default": false,
+          "description": "Include additional information such as request URL and response time"
```
Huachao: also indicate that this only works when previewing response in untitled document

Huachao:
@dakaraphi Merged, I will publish this change in next release. Thanks for your contribution 😄

Huachao:
@Download @dakaraphi I have published the new version 0.20.3 with this new `setTextDocumentLanguage` API

PR #89, opened 2017-04-06
Switch to new API for opening editor

dakaraphi:
This is a change to the open response in Untitled document.
VSCode 1.11 has a new parameter on the API which allows to set the content on opening.

I think this is better for these reasons:
1)  This allows the document to open without a previous undo.  So if you hit undo, you will no longer get a blank document
2)  Since the document is opened in a state of not dirty, you can close the document without the editor giving you a popup asking if you want to close the document.

@Huachao Let me know what you think

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "url": "https://github.com/Huachao/vscode-restclient.git"
   },
   "engines": {
-    "vscode": "^1.9.0"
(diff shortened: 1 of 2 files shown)


Huachao:
@dakaraphi Merged, thanks. Really nice change.

PR #93, opened 2017-05-05
use new vscode api to preview untitled document

dakaraphi:
vscode 1.12 added a new `showTextDocument` API.

Using this API, it is now possible to reuse the same untitled tab and set the language.

### DIFF
--- a/package.json
+++ b/package.json
@@ -411,7 +411,7 @@
     "mocha": "^3.2.0",
     "tslint": "^4.2.0",
     "typescript": "^2.0.3",
-    "vscode": "^1.1.0"
+    "vscode": "^1.0.0"
   },
   "dependencies": {
     "applicationinsights": "^0.18.0",
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "url": "https://github.com/Huachao/vscode-restclient.git"
   },
   "engines": {
-    "vscode": "^1.11.0"
+    "vscode": "^1.12.0"
   },
   "categories": [
     "Other",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -411,7 +411,7 @@
     "mocha": "^3.2.0",
     "tslint": "^4.2.0",
     "typescript": "^2.0.3",
-    "vscode": "^1.0.0"
+    "vscode": "^1.1.0"
```
Huachao: why do we need to upgrade this?

dakaraphi:
Ahh, I forgot which entry fetches the typings dependency and then forgot to change it back.
Fixed now, thanks.

src/views/responseUntitledFileContentProvider.ts
```diff
@@ -8,33 +8,12 @@ import { EOL } from 'os';
...
-                return;
-            }
-        }
 
+        const language = autoSetLanguage ? UntitledFileContentProvider.languageFromContentType(response) : 'http';
         const content = UntitledFileContentProvider.formatResponse(response, language, additionalInfo, autoSetLanguage);
         workspace.openTextDocument({ 'language': language, 'content': content }).then(document => {
             UntitledFileContentProvider.createdFiles.push(document);
-            window.showTextDocument(document, ViewColumn.Two, false);
+            window.showTextDocument(document, {viewColumn: ViewColumn.Two, preserveFocus: false, preview: !createNewFile}) ;
```
Huachao: nice change, but one thing I found is that, if we don't create new file(or tab), the existing tab title will still be changed for each request from Untitled-1 to Untitled-N. Do you have any ideas?

dakaraphi: Yes, I noticed that too.  However, I expect this is actually intentional by vscode.
Imagine that you are previewing documents which may be nearly identical.
By changing the number each time, you have better visual feedback that the document actually changed.

Huachao: Make sense, thanks @dakaraphi

Huachao:
@dakaraphi Merged, thanks

Issue #74, opened 2017-02-27
Provide possibility to extract values from response


akalcik:
It will be helpful to have a possibility to extract items from response into the variables. For example I want test API which is protected by Authorization Server. Today, to access the API, I need start request to Authorization Server, copy the claim from response into the new request and start new API request. It will be nice if I can extract the claim into the variable.

Huachao:
@akalcik nice suggestion, I will consider it carefully 😄

pfemiani:
It would be really helpful to get values from responses. One idea could be to allow naming of Requests, and a named request would result in its own Response tab/doc (rather than trying to manage response variables/scopes/etc), and you could perhaps pull values via {{RequestName:ResponseValueLocation}} where response value location might be in "dot" notation for Json or XPath for Xml responses?

```
### Auth
POST https://auth.server.com/auth HTTP/1.1
Authorization: BLAH My:Creds
Accept: application/json

### MyInfo
GET https://api.server.com/me HTTP/1.1
Authorization: TOKEN {{Auth:data.AuthToken}}
Accept: application/json
```

Theoretically you could then infer request dependency, so if Auth response doesn't exist yet it could execute automatically before the MyInfo request, but that is really just a nice-to-have.

pcornier:
It would be a great improvement 🤘, I really want to get rid of Postman in my workflows.

fmartingr:
I'm also looking forward to this. While I agree with @pfemiani proposal, a simpler approach could be to allow saving response to a variable and then using it:

```
POST url/auth -> authResponse
Content-Type: application/json

{...}

###

POST my/protected/endpoint
Authorization: Token {{myResponse.json.token}}
```

I'm not sure how other rest client for other editors resolve this (if they do at all) but it will be an awesome addition.

lee-fu:
This would be an awesome addition. I would like to assert against the response as a sort of basic contract test like so:
```
POST ...
...
const expected = {...};
assert({response} === expected, 'The response should equal the expected payload');
```

cbrevik:
Hi @Huachao!

Love this extension, use it daily, and I would love this feature as well. 

I tried to do a really naive proof-of-concept implementation of this here: https://github.com/cbrevik/vscode-restclient/commit/e716a00b78e62eeb68a1a81fadcb622fc87646da

It's inspired by @fmartingr's suggestion of "piping" the response to a variable name like `POST url/auth -> authResponse`.

That implementation should work with something like:
```rest
[...]

PR #140, opened 2017-12-03
Provide possibility to extract values from response

cbrevik:
As stated in https://github.com/Huachao/vscode-restclient/issues/74#issuecomment-348782450, this is a naive proof-of-concept implementation of #74.

It's inspired by @fmartingr's suggestion of "piping" the response to a variable name like;
`POST url/auth -> authResponse`.

The goal would be to allow the developer refer to the response by a variable name, and use a "dotted" property approach to access its values.

This implementation should work with something like:
```rest
@baseUrl = https://jsonplaceholder.typicode.com
GET {{baseUrl}}/posts/1 -> myResponse

###

GET {{baseUrl}}/users/{{myResponse.body.userId}}
```

Opening PR for discussion of approach. I do not consider this to be a ready solution yet. Would love input from @Huachao.

Things to discuss:
- How to parse `response.body` (and possibly other [`HttpResponse`](https://github.com/Huachao/vscode-restclient/blob/32657c889f0ed134b8f889820360211f522cc5de/src/models/httpResponse.ts) values?). This naive approach does a `JSON.parse` automatically on `body`, which is not satisfactory. Maybe adding `jsonBody`, `xmlBody`, etc helper properties like [@fmartingr suggested](https://github.com/Huachao/vscode-restclient/issues/74#issuecomment-348787063) is one approach.
- This approach means the response variable can be accessed across files, because it is saved statically on `ResponseStore`. Should it be tied to the file which ran the request instead? Or is there value in cross-file sharing? Can be confusing if variables start overwriting each other on name collisions.
- Possible to support autocomplete like with other variables? Dotted autocomplete?
- Other things I haven't thought of?


Huachao:
@cbrevik @fmartingr @akalcik thank you all, and @cbrevik, this is a really nice patch that make a solid step for implementing this feature. And after we finalized the `Chaining Request` design, we can continue work on this PR to make this all ready, before that I still have some questions to discuss with you.

1.  Since we need to have a syntax to identify the response, what's the proper syntax. And you've provided a cool way to use __`->`__, like `GET https://api.example.com -> exampleResp`, or define request name in comment to reference later.
```http
###
# @name ExampleRequest
GET https://api.example.com

###
``` 
2. For a already defined request/response name, what's the scope of the name, is it file scope or global scope? That means, for the former, if we define a response named `exampleResp`, we can't use it in other request files. 
3. What's the life time of the response? If we store them in cache(memory), when we restart _vscode_, these responses won't be available, and those requests which reference other requests won't succeed.  Or we need to persist the response to disk.
4. I'd like to provide several response reference types as you do, like _Raw Body_, _Body Elements_(with syntax similar to __XPath__/__JSONPath__) and _Response Header Value_. Do you think it's enough.

Again, thank you all for your great contribution to this extension. 😄

akalcik:
@Huachao @cbrevik @fmartingr

we should go agile and try to make simple and usable solution, which can be enhanced later.

1. I'm fine with the syntax, we can use ```|``` instead of ```->```, but no really matters.
2. Ok with file scope
3. Ok with persistence in the memory
4.  Yes

When we are working with the JSON it will be probably helpful to use same JSON Query Language, like https://marketplace.visualstudio.com/items?itemName=joshwong.vscode-jmespath or http://www.jsoniq.org/

cbrevik:
1. It might be that @Huachao's approach here (adding a `@name` tag in a comment `#` is a good idea. The problem with adding `->` is that it is not exactly HTTP standard compliant. Also doing it with a tags means we could maybe add additional functionality. Say we want the `response.body` deserialized from JSON could be done by adding a `@json` tag. I.e.
    ```http
    # @json @name ExampleRequest
    GET https://api.example.com
    ```
    Might also be more extensible, `@xml`, etc. Thoughts?
2. File scope should probably be the default. But again, if we use "tags" we could use something like `@global` or whatever to give it global scope. Maybe sometimes we want to overwrite/use a variable cross-file?
3. I agree that there might be a need to cache stuff on disk at some point, but that is maybe out-of-scope for this feature? I think in-memory is good enough for now.
4. Is there a need to go all-out XPath? Since both properties and array-indices can be accessed through brackets (`[ ]`), maybe `response.body.someItemArray.0.itemId` dotted syntax is okay? Seems like XPath at least has a heavy and often misunderstood syntax.

fmartingr:
1) Having the variable name in a comment before it would make for future request configuration for more stuff and also make the file compatible with other editor's rest-client plugins, so 👍 
2) File scope is enough IMO.
3) I think that using memory is enough too, if you're trying to do a request that references another one that you did not run, the request could either fail silently or throw an error explaining to the user that a previous request is required.
4) In most scenarios the user would only want the serialized body to use response parameters in another request, we just need to identify what content-types to support first (json/xml?)

I'm not denying/ignoring the comments for more enhancements here but I'm focusing only to reach a *MVP* feature set where this is usable and when is ready and launched iterate more over it.

Huachao:
@fmartingr @cbrevik @akalcik thank you all for your quick responses. And I summarize the answers for developing this feature:

1.  Use comment as the syntax to define referenced request, e.g.,
```http
# @name login
POST https://api.example.com/login HTTP/1.1
Content-Type: application/json

{
   "user":  "xxx",
   "password": "yyy"
}

###

GET https://api.example.com/docs HTTP/1.1
Authorization: Bearer {{login.resp.headers.token}}
```
2. Set request name to file scope, document uri needs to be part of cache key.
3. Use in memory cache to store referenced response, and if the referenced response is not found in memory(e.g., restart vscode), display a warning message box.
4. Judge response content type(JSON/XML) from response `Content-Type` header, and use JSONPath/XPath syntax for each case. And currently only these two formats are supported.

Do you agree with this design, and any other suggestions? 😊

cbrevik:
Agreed on all points :) But I'm still wondering about number 4. When you say XPath syntax, do you mean full syntax like [this](https://www.w3schools.com/xml/xpath_syntax.asp) with node/attribute retrieval, predicates, etc. Or just a simple `root/someNode[0]/childNode[1]`. 

Is full syntax easy to support in JavaScript?

Huachao:
@cbrevik I can't agree with you more. And maybe we can first support JSON path only, and leave the XPATH(or related solution) later.

fmartingr:
I agree with that as first iteration. I don't know how complex will be the XPath for XML because I haven't developed a lot with JS/TS but if it comes bundled in it could be nice to have at least a working way to use them.

Have any of the people from #140 mentioned they want this for XML explicitly? Maybe we are trying to support it when no one has requested it (yet).

cbrevik:
I don't think XML support is strictly necessary as an MVP either. 

I've done a preliminary implementation at https://github.com/cbrevik/vscode-restclient/commit/f51bc85f04321c81b30b7c6845ff9c60f1bdc479

Not finished of course. I'll merge into this PR (or create a new PR) when it is working better. And we can iterate from there.

Huachao:
@cbrevik thanks so much, and waiting for your updated PR.

akalcik:
Actually I liked the first proposal of @cbrevik much more!  Defining variables in comment seems to be very confusing. What about this one? This is pretty clear and aligned with current syntax.

```
@myResponse
@baseUrl = https://jsonplaceholder.typicode.com
GET {{baseUrl}}/posts/1 -> @myResponse
```

```@myResponse``` is in this case own declared variable. We can introduce later also convention based variables (for example ```@jsonMyResponse```) for variables which content will be deserialized automagically.

```
@myResponse
@jsonMyResponse
@baseUrl = https://jsonplaceholder.typicode.com
GET {{baseUrl}}/posts/1 -> @myResponse | @jsonMyResponse
```

@fmartingr @Huachao 

_"...file compatible with other editor's rest-client plugins, so..."_

Just confused. Is the file compatible now? I think to make file compatible with other editor's we need to redesign the syntax completely.

**Comment by cbrevik** [comment]
@akalcik The `->` is sort of invented syntax, which might be even more confusing for people. 
Of course the variables `{{variable}}` is "invented" syntax which is non-valid as well. But they are typical of "merge fields", think `handlebar.js`, etc.
Using comment-based `@tags` seems a bit cleaner to me, and more extensible.

@Huachao I've pushed a few more changes to this PR, reverting the old commit. It's working in a very basic manner, with red diagnostic squiggly lines if variable reference is not in memory. 
Basic code completion of response variables, basic hover with printed value.

Just thought I'd check in here. Not done yet. Needs more error handling, better `response.body` parsing, and will refactor so it's a bit cleaner.

**Comment by Huachao** [comment]
@akalcik sorry for the late to response to your comment about response declaration syntax.

@akalcik @cbrevik @fmartingr I'd like to share with you about my consideration for three most requested feature(#67 #74 #86), I think them can all benefit from this PR.

I have several reasons about why personally I suggest to use a syntax in comment not in request line.
As @cbrevik mentioned, use `@tag` in comment seems more extensible and cleaner to me. I think what the syntax represents is not just define response variable, it's more of a way to add metadata to the underling request. So with this syntax we can have more powerful capabilities, like supporting API request documentation(#67), response assertion/validation.

And a key difference of our proposals is that I use request name instead of response name to reference the response. Many of existing REST API tools(e.g., _Postman_, _Insomnia_) have the ability to add name for the request, so this way has a good compatibility if we later want to import/export data from those tools. And we even have the ability to reference other request's part, not only response. And when user references other request/response with name, we can also easily know the request dependency, even we can trigger dependent request automatically.

Another reason that I don't want to use `->` to represent creating a new variable is that I just want to create a file scope variable with original syntax: `@variableName: variableValue`. Since it's more explicit to know what the variable value should be. So when I am considering another requested feature(#86), I can also leverage from this feature to store token in variable automatically with little cost, like following:
```http
# @name login
POST https://api.example.com/login
Content-Type: application/json

{
  "user_name": "foo",
  "password": "bar"
}

###

@token = {{login.resp.body.access_token}}

...
```

Below is a full example `http` file for this issue:
```http
# @name listDocs
# @descirption A test request to list all the docs
GET https://api.example.com/docs
Authorization: token 123456

###

# @name createDoc
# @status_code 201
# @resp_content_type application/json
POST https://api.example.com/docs
Authorization: {{listDocs.req.headers.Authorization}}
Content-Type: application/json

{
  "author": "foo",
  "content": "bar"
}

###

# @name getDoc
GET https://api.example.com/docs/{{createDoc.resp.body.id}}

```

akalcik:
@Huachao @cbrevik I get your points. Still confused to understand some language design decisions. Please help me.

When is a file scope variable created it is like this:

```@variableName = variableValue```

Name of this variable is a ```variableName ```  and value is ```variableValue```,  scope is a file. Actually no brainer. We can call this type of variable a _constant_ as the value can't be change again.

Now:

```
# @name listDocs
# @description A test request to list all the docs
GET https://api.example.com/docs
Authorization: token 123456
```
In this case the we have a two variables with name ```name``` and ```description```. Scope of those variables seems to be also file. Now the most confusing part for me ist following. ```listDocs``` seems to be a somehow placeholder for the request,  which can be used later. I don't really know how I should name it properly. Now what is a value of ```description``` variable? ```A test request to list all the docs```, right? In one case value of the variable is a placeholder,  in other case seems to be specific value. Not sure if the value of ```listDocs``` and ```description``` can be changed later. My question is: what define what can I use as a value? It is by a convention? When it is a convention then I understand. I would call than the variables like request scoped variables and it shouldn't be possible to change the values later. When it is convention than I would propose to change ```name``` to ```request```. 

What  also I would propose to change is the missing operator. In one case is value set per ```=``` in other case act ```  ``` (empty space) as a operator.

**Comment by Huachao** [comment]
@akalcik the `@tag` defined before each request in comments, I'd like to call it metadata not variable. So in this way, you don't create two variables named `name` and `description`, just add two metadata for underlying request, one is `name` which can be used as an identifier of the request to reference, another is `description` which may used to add some description of the request(which can be later export to other REST tools). You can think of them as `attributes` of the _method_ compared with `C#`(I guess maybe you are familiar with). 

So the metadata names are predefined for special use, currently seems we only need the `name` metadata for this feature, since the `name` metadata are used as identifier of the request, if not defined this metadata, I think the request is an anonymous request, which can't be referenced by other requests. And the __value__ of these metadata can be changed as your wish, just remind that, __name__ metadata is special since it's an identifier of the request and can be used by other requests, so if you update the name, you also need to update the reference parts(that means the value of @name is __file scope__).

And the syntax of adding metadata to request just brought from _JSDoc_, I don't have many good ideas either, @cbrevik @akalcik @fmartingr what kind of syntax do you think appropriate for this?

Thanks again @akalcik @cbrevik @fmartingr , without your discussions, I can't think about it deeply.

**Comment by narayanan** [comment]
Good Discussion. How about using `!` as the meta character to declare the metadata?. This will also allow declaring metadata out side a comment.

`!name` - mechanism to declare a file scope variable of type Javscript object which can be used to access various aspects of the request / response
`!description` - mechanism to describe the purpose of the API invocation

Example

```
!name @listDocs
!descirption "A test request to list all the docs"
GET https://api.example.com/docs
Authorization: token 123456

###

!name @createDoc
!status_code 201
!resp_content_type "application/json"
POST https://api.example.com/docs
Authorization: {{listDocs.req.headers.Authorization}}
Content-Type: application/json

{
  "author": "foo",
  "content": "bar"
}

###

!name getDoc
GET https://api.example.com/docs/{{createDoc.resp.body.id}}
```

**Comment by Huachao** [comment]
@cbrevik any updates on this, sorry to bother you.

**Comment by cbrevik** [comment]
@Huachao I've been extremely busy since Christmas. So I haven't had a chance to work on this, sorry. I'll try to devote some more time this weekend. 

It's basically in a working state. The difference from your proposal is that I've only cached the `Response`. So it needs to be updated to cache the `Request` as well. Other than that, my previous comment stands about needing better error handling, refactoring, etc.

**Comment by Huachao** [comment]
@cbrevik that's ok, I am sorry that I didn't intend to push you, and I've fixed some conflicts and pushed into your branch.

**Comment by lostfields** [comment]
I really like this, when will it hit marketplace?

For xml and other strange format, you could support regex as well, lets say you have the captured data in myDoc after using #@name myDoc

In a sub-sequent request you may define variable `@mycapture = /(\b\w+\b)/.exec({{myDoc}})[1]` to capture the first word in response myDoc. This will make it support any responses as text, xml and json.

**Comment by lostfields** [comment]
and one more thing, maybe you never should use `{{createDoc.resp.body.id}}` in your request directly, maybe you always should assign it to a variable `@mycapture = {{createDoc.resp.body.id}}`. (combine this with regex extracting as well). 

If you want to capture the hole content into `createDoc` I really liked the syntax of a previous response here;

`
@user = 1
@createDoc
POST ../users/{{user}}/invoices_ HTTP/1.1`

The response is then captured and assigned to @createDoc, following normal variable syntax, except that it does not have a equal sign making the next line to be the content.

**Comment by aphilpotts** [comment]
Just a thumbs up for anything, and I won't quibble about the syntax, that can set some variable sourced from a previous JSON response

Our APIs always return JSON (and say so) and tend to return a lot of different ids, not limited to Bearer tokens, it is not ideal to copy paste all the time.

Can this handle the case of a variable _inside an array_? Something like 

@firstId = "{{response.body.foo.bar[0].id}}" ?

I know this is a slippery slope to a fully fledged scripting language, so the feature creep has to stop somewhere!

**Comment by Huachao** [comment]
@aphilpotts JSONPath of response body can be supported I think. Is that right, @cbrevik ? If so, your requirements can be met

**Comment by cbrevik** [comment]
I'm implementing something similar to JSONPath, and yes it should support arrays lookup.

Not sure if there is support for re-assignment of a variable though. That is to say `@firstid = "{{response.body.foo.bar[0].id}}" `. 

Surely this is not supported today?:
```
@someVar = Hello
@reassignedVar = {{someVar}}
```

`{{reassignedVar}}` here will be literally the text `{{someVar}}`?

**Comment by Huachao** [comment]
@cbrevik this is not supported today, I just think about the possibility in the future

**Comment by aphilpotts** [comment]
Yes, this will work for me, thanks!

**Comment by aphilpotts** [comment]
Ah, just due to me not understanding! My syntax was made up, and I was really focusing on the part regarding referencing into arrays. That has been answered.

As long is there is some way to name the referenced value, so it  can be used in a number of other places, is all I was trying to express.

I would not want to embed a complex reference, like {{response.body.foo.bar[0].id}} in a number of other requests

**Comment by cbrevik** [comment]
@Huachao I've done quite a bit of refactoring now to also cache the request, along with the response. Basically I've introduced the concept of a `RequestVariable`.

I've squashed the bugs I've found. There are probably more. But I consider the implementation a working first draft. 

I've implemented code completion for the variables, hover over each part for value, and diagnostic (red squiggly lines) if the variable is not loaded in memory.

Maybe you could test out my implementation a bit and maybe review the code?

File: src/variableProcessor.ts [hunk 1]
```diff
@@ -48,6 +54,23 @@ export class VariableProcessor {
...
+        for (let [variableName, variableValue] of requestVariables) {
+            let regex = new RegExp(`\\{\\{\\s*${variableName}.*\\s*\\}\\}`, 'g');
+            let matches = request.match(regex);
+            if (matches && matches.length > 0) {
+                for (var i = 0; i < matches.length; i++) {
+                    var requestVariable = matches[i].replace('{{', '').replace('}}', '');
+                    try {
+                        const value = RequestVariableCacheValueProcessor.getValueAtPath(variableValue, requestVariable);
+                        request = request.replace(new RegExp(`\\{\\{\\s*${requestVariable}\\s*\\}\\}`, 'g'), value.toString());
+                    } catch {
```
Huachao: Does catch without exception work in your environment? I thought JS needs us to provide an error variable even if we don't need it like
```js
catch (e) {
...
}
```

According to this [doc](https://docs.microsoft.com/en-us/scripting/javascript/reference/try-dot-dot-dot-catch-dot-dot-dot-finally-statement-javascript), it explicitly said it's required

src/requestVariableCacheValueProcessor.ts
```diff
@@ -0,0 +1,107 @@
...
+        const {body} = response;
+        const contentType = response.getResponseHeaderValue("content-type")
+        if (contentType) {
+            let mime = MimeUtility.parse(contentType);
+            let type = mime.type;
+            let suffix = mime.suffix;
+            if (type === 'application/json' || suffix === '+json') {
+                try {
+                    return JSON.parse(body);
+                } catch {
```
Huachao: same as below

src/variableProcessor.ts
cbrevik: Well, it works in my dev environment because I normally work with TypeScript 2.5+, which actually [does support optional catch clauses](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#typescript-25)

I see this repository depends upon `"typescript": "^2.0.3"`, but the caret `^` symbol there means that it will match any `2.x.x` release, up till but not including `3.0.0`. So that's probably why it works on my machine.

I'd recommend upgrading the `typescript` version anyway, there's a lot of nice stuff in the newer version. Also recommend locking down the actual version instead of using carets `^` or tilde `~`.

src/requestVariableCacheValueProcessor.ts
cbrevik: See above.

cbrevik:
Totally forgot about `tslint`, sorry! I normally use https://github.com/prettier/prettier day-to-day which is an auto-formatter instead. I find it to be a lot less work than have to manually fix linting issues.

Anyway I'll fix the styling issues! I'd recommend having a look at `prettier` though!

src/variableProcessor.ts
Huachao: @cbrevik Nice, I haven't known this before, nice, please ignore this comment.

src/constants.ts
```diff
@@ -60,4 +60,6 @@ export const AzureClouds = {
 
 export const CommentIdentifiersRegex: RegExp = /^\s*(#|\/{2})/;
 
-export const VariableDefinitionRegex: RegExp = /^\s*@([^\s=]+)\s*=\s*(.+)\s*$/;
\ No newline at end of file
+export const VariableDefinitionRegex: RegExp = /^\s*@([^\s=]+)\s*=\s*(.+)\s*$/;
+
+export const RequestVariableDefinitionRegex: RegExp = /#.*\s+@name\s+(\w+).*/;
```
Huachao: I think the request variable definition should occupy a whole line, so I think the syntax may be like `^\s*#{1,}\s+@name\s+(\S+)\s*$`. One important difference between our proposals is that no non whitespace characters are allowed between __#__ and __@name__

src/variableProcessor.ts
```diff
@@ -1,14 +1,19 @@
 'use strict';
+import { VariableType } from "./models/variableType";
```
Huachao: Adjust the import order, move them in regions of custom import

In my code structure, the imports are sorted as below:
1. VS Code imports
2. Custom imports (From our extension internal code)
3. 3rd party libraries import

So could you please resort all your files? Thanks in advance.

src/variableProcessor.ts
```diff
@@ -48,6 +53,23 @@ export class VariableProcessor {
             }
         }
 
+        let requestVariables = VariableProcessor.getRequestVariablesInCurrentFile();
+        for (let [variableName, variableValue] of requestVariables) {
+            let regex = new RegExp(`\\{\\{\\s*${variableName}.*\\s*\\}\\}`, 'g');
+            let matches = request.match(regex);
+            if (matches && matches.length > 0) {
+                for (let i = 0; i < matches.length; i++) {
+                    const requestVariable = matches[i].replace('{{', '').replace('}}', '');
```
Huachao: should also remove whitespaces after `{{` and before `}}`. So it should be something like
```js
const requestVariable = match[i].replace(/\{\{\s*/, '').replace(/\s*\}\}/, '');
```

src/httpCodeLensProvider.ts
```diff
@@ -39,9 +39,14 @@ export class HttpCodeLensProvider implements CodeLensProvider {
             }
 
             if (blockStart <= blockEnd) {
+                let requestVariable = null;
+                if (blockStart > 0) {
```
Huachao: This seems will break following `http` request if user has other comments after the request variable definition line:

```http
# @name login
# This is a test request!
POST https://api.example.com/login
...
```

src/httpCodeLensProvider.ts
```diff
@@ -39,9 +39,14 @@ export class HttpCodeLensProvider implements CodeLensProvider {
...
+                    requestVariable = Selector.getRequestVariableDefinitionName(lines[blockStart - 1]);
+                }
                 const range = new Range(blockStart, 0, blockEnd, 0);
+                const requestLines = new RequestLines(range, document.uri.toString(), requestVariable);
```
Huachao: This seems you only add `requestVariable` when user trigger via the code lens action, if not, like using shortcut or _F1 Command_, no variables will be available. Since my extension also allows user just needs to click the mouse in the expected request range, and then trigger request via shortcut and _F1 command_

src/models/requestVariableCacheValue.ts
```diff
@@ -0,0 +1,11 @@
+"use strict";
+
+import { HttpRequest } from "./httpRequest";
+import { HttpResponse } from "./httpResponse";
+
+export class RequestVariableCacheValue {
+    public constructor(
+        public request: HttpRequest,
+        public response?: HttpResponse) {
```
Huachao: Why add a `?` here, when response doesn't meet?

src/controllers/requestController.ts
```diff
@@ -135,6 +142,9 @@ export class RequestController {
 
             let previewUri = this.generatePreviewUri();
             ResponseStore.add(previewUri.toString(), response);
+            if (httpRequest.requestVariableCacheKey) {
```
Huachao: If the response is not successful according to the response status code, do we really need to save the response to cache? In other words, does the `RequestVariableCache` only saves the successful response.

aphilpotts: I don't know if this contributes, but I have encountered cases where extra information is returned in the content, for say a 404

Huachao: OK, @aphilpotts your advice is we sho uld keep existing logic that in some cases we need to reference the content in the error response

src/variableProcessor.ts
cbrevik: Is it a good idea to allow whitespace after `{{` and before `}}` in variables? Is it a general requirement?

Then I cannot re-use utilities like `isVariableReferenceFromLine` in `variableUtility.ts`. In fact that will require a bit of re-write to support hover & completion.

src/models/requestVariableCacheValue.ts
cbrevik: The idea at first was; cache request as it is run, then update cache with response when it is returned. But I dropped the idea since the use of the variable sort of depends on the response completing. Forgot to revert the `?`, will do now.

src/controllers/requestController.ts
cbrevik: Yes I've encountered cases where 4xx/5xx still returns content. Might actually be cases where people want to retrieve headers from error responses as well. I'd prefer to keep as-is.

src/httpCodeLensProvider.ts
cbrevik: Will look into this

src/httpCodeLensProvider.ts
cbrevik: Rewrote the implementation so it instead checks within the delimited text scope of the request being run.

src/httpCodeLensProvider.ts
cbrevik: Should work both with F1 Commands, Code Lens and selecting text now.

src/variableProcessor.ts
```diff
@@ -1,14 +1,19 @@
 'use strict';
 
-import { commands, QuickPickItem, QuickPickOptions, Uri, window } from 'vscode';
+import { commands, QuickPickItem, QuickPickOptions, Uri, window, TextDocument } from 'vscode';
 import { EnvironmentController } from './controllers/environmentController';
-import * as Constants from './constants';
+import * as Constants from "./constants";
 import { Func } from './common/delegates';
+import * as moment from "moment";
```
Huachao: Move import statement  to the original place please.

src/variableProcessor.ts
Huachao: Agree, keep this logic.

src/requestVariableCacheValueProcessor.ts
```diff
@@ -0,0 +1,109 @@
...
+
+export class RequestVariableCacheValueProcessor {
+    public static getValueAtPath(value: RequestVariableCacheValue, path: string): any {
+        if (!value
+            || !path
+            || !RequestVariableCacheValueProcessor.isValidPath(path)) {
+            return undefined;
+        }
+
+        const parts = path.split(".");
```
Huachao: Can we simply use array destruction syntax here, which seems more readable than `parts` array, i.e.
```js
const [name, source, type, ...rest] = path.split(".");
```

src/requestVariableCacheValueProcessor.ts
```diff
@@ -0,0 +1,109 @@
...
+            default:
+                return undefined;
+        }
+    }
+
+    private static resolveResponse(response: HttpResponse, parts: string[]) {
+        if (parts.length >= 3) {
+            const matches = parts[2].match(partRegex);
+            // Must parse body specifically
+            if (matches[1] === "body") {
```
Huachao: Would you like to support header too?

src/requestVariableCacheValueProcessor.ts
```diff
@@ -0,0 +1,109 @@
+"use strict";
+import { window } from "vscode";
+
+import { MimeUtility } from "./mimeUtility";
+import { RequestVariableCacheValue } from './models/requestVariableCacheValue';
+import { HttpResponse } from './models/httpResponse';
+
+const validPathRegex = new RegExp(/(\w+)(\.\w+|\[\d+\])*/);
```
Huachao: Can we move this syntax check regex directly into the regex `let regex = new RegExp(`\\{\\{\\s*${variableName}(.*)\\s*\\}\\}`, 'g')` of **variableProcessor.ts**, so that we only allow matched request variables to flow into `getValueAtPath ` logic.

src/requestVariableCacheValueProcessor.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    private static resolveResponse(response: HttpResponse, parts: string[]) {
+        if (parts.length >= 3) {
+            const matches = parts[2].match(partRegex);
+            // Must parse body specifically
+            if (matches[1] === "body") {
+                let bodyValue = RequestVariableCacheValueProcessor.parseResponseBody(response);
+                // If array indeces:
+                for (let j = 2; j < matches.length; j++) {
```
Huachao: Do we support json path here, if so, can we use some existing packages on npm to do such kind of parsing/querying job?

src/models/httpRequest.ts
```diff
@@ -1,14 +1,16 @@
...
 
 export class HttpRequest {
     public constructor(
         public method: string,
         public url: string,
         public headers: { [key: string]: string },
         public body: string | Stream,
-        public rawBody: string) {
+        public rawBody: string,
+        public requestVariableCacheKey?: RequestVariableCacheKey) {
```
Huachao: It seems we won't pass this in `constructor` call, how about move this to a class instance field, which should be supported by TypeScript:
```ts
export class HttpRequest {
    public requestVariableCacheKey?: RequestVariableCacheKey;
    public constructor(
        public method: string,
        public url: string,
        public headers: { [key: string]: string },
        public body: string | Stream,
        public rawBody: string) {
    }
```

src/selector.ts
```diff
@@ -33,6 +44,11 @@ export class Selector {
         return rows;
     }
 
+    public static getRequestVariableDefinitionName(text: string): string {
+        const matched = text.match(Constants.RequestVariableDefinitionRegex);
+        return matched && matched.length >= 2 ? matched[1] : null;
```
Huachao: directly matched[1] is enough I think

src/variableDiagnosticsProvider.ts
```diff
@@ -0,0 +1,107 @@
...
+
+        existArray.forEach((ea) => {
+            if (!ea.exists) {
+                vars.forEach((v) => {
+                    if (v.variableName === ea.name) {
+                        diagnostics.push({
+                            severity: DiagnosticSeverity.Error,
+                            range: new Range(new Position(v.lineNumber, v.startIndex), new Position(v.lineNumber, v.endIndex)),
+                            message: `${v.variableName} is not loaded in memory`,
+                            source: 'ex',
```
Huachao: Change `ex` to `http` is more understandable to user.

src/variableDiagnosticsProvider.ts
```diff
@@ -0,0 +1,107 @@
...
+                            code: "10",
```
Huachao: Do we really need code here, it seems this is only used for CodeActionContext

src/variableDiagnosticsProvider.ts
```diff
@@ -0,0 +1,107 @@
...
+
+import { event } from "./events/requestVariableEvent";
+import { VariableProcessor } from "./variableProcessor";
+
+export class VariableDiagnosticsProvider {
+
+    private diagnosticCollection: DiagnosticCollection;
+
+    public activate(subscriptions: Disposable[]) {
+        this.diagnosticCollection = languages.createDiagnosticCollection();
```
Huachao: minor: add a diagnostic collection name like `http`

src/variableDiagnosticsProvider.ts
```diff
@@ -0,0 +1,107 @@
...
+import { workspace, languages, Diagnostic, DiagnosticSeverity, DiagnosticCollection, TextDocument, Range, Position, Disposable } from 'vscode';
+
+import { event } from "./events/requestVariableEvent";
+import { VariableProcessor } from "./variableProcessor";
+
+export class VariableDiagnosticsProvider {
+
+    private diagnosticCollection: DiagnosticCollection;
+
+    public activate(subscriptions: Disposable[]) {
```
Huachao: Is it possible to move the language diagnostic collection create and register text document related listeners in `extension.ts`. Since all of our language providers registered there. It will be easier to manage for organizing in one single place.

src/variableDiagnosticsProvider.ts
```diff
@@ -0,0 +1,107 @@
+'use strict';
+
+import { workspace, languages, Diagnostic, DiagnosticSeverity, DiagnosticCollection, TextDocument, Range, Position, Disposable } from 'vscode';
+
+import { event } from "./events/requestVariableEvent";
```
Huachao: How about doing a import rename like `import { event as OnRequestVariableEvent } from "./events/requestVariableEvent";` I think it's more readable than `event`

src/variableDiagnosticsProvider.ts
```diff
@@ -0,0 +1,107 @@
...
+            }
+        });
+
+        this.diagnosticCollection.set(document.uri, diagnostics);
+    }
+
+    private findVariables(document: TextDocument) : Variable[] {
+        let vars: Variable[] = [];
+        let lines = document.getText().split(/\r?\n/g);
+        let pattern = /\{\{(\w+)(\.\w+|\[\d+\])*\}\}/;
```
Huachao: This seems too strict(e.g., `-`, `_` not allowed), and not consistent with `RequestVariableDefinitionRegex` in `constants.ts`, I think `\S` is OK. Do you think so?

src/variableDiagnosticsProvider.ts
```diff
@@ -0,0 +1,107 @@
...
+    private async checkVariables(document: TextDocument) {
+        if (document.languageId !== 'http') {
+            return;
+        }
+
+        let diagnostics: Diagnostic[] = [];
+
+        let vars = this.findVariables(document);
+
+        let varNames = vars.map((v) => v.variableName.split(".")[0].split("[")[0]);
```
Huachao: Why do we need an extra split for `[`?

src/variableProcessor.ts
```diff
@@ -366,4 +394,74 @@ export class VariableProcessor {
...
+        let requestVariables = VariableProcessor.getRequestVariablesInFile(document);
+
+        let variableDefinitions = new Map<string, VariableType[]>();
+        fileVariables.forEach((val, key) => {
+            variableDefinitions.set(key, [ VariableType.Custom ]);
+        });
+
+        environmentVariables.forEach((val, key) => {
+            if (variableDefinitions.has(key)) {
+                let types = variableDefinitions.get(key);
```
Huachao: Seems one line is enough:
```js
variableDefinitions.get(key).push(VariableType.Environment);
```

src/variableDiagnosticsProvider.ts
```diff
@@ -0,0 +1,107 @@
...
+
+        // Distinct varNames
+        varNames = Array.from(new Set(varNames));
+
+        let existArray = await VariableProcessor.checkVariableDefinitionExists(document, varNames);
+
+        existArray.forEach((ea) => {
```
Huachao: Here I prefer object destructing syntax `{name, exists}`

src/events/requestVariableEvent.ts
```diff
@@ -0,0 +1,13 @@
...
+import { RequestVariableCacheKey } from "../models/requestVariableCacheKey";
+
+interface RequestVariableEvent {
+  cacheKey: RequestVariableCacheKey;
+}
+
+const eventEmitter = new EventEmitter<RequestVariableEvent>();
+
+export const fireEvent = (event: RequestVariableEvent) =>
+  eventEmitter.fire(event);
```
Huachao: Do we really need the `RequestVariableEvent`, if so, I think we should also export the `RequestVariableEvent`, otherwise, no one can use it in their own listeners.

src/variableDiagnosticsProvider.ts
```diff
@@ -0,0 +1,107 @@
...
+        workspace.onDidOpenTextDocument(this.checkVariables, this, subscriptions);
+        workspace.onDidCloseTextDocument((textDocument) => {
+            this.diagnosticCollection.delete(textDocument.uri);
+        }, null, subscriptions);
+        workspace.onDidSaveTextDocument(this.checkVariables, this, subscriptions);
+
+        // Check all open documents
+        this.checkVariablesInAllTextDocuments();
+
+        event(() => {
```
Huachao: No arrow function is needed, just like `event(this.checkVariablesInAllTextDocuments);` is OK

src/variableUtility.ts
```diff
@@ -20,6 +20,22 @@ export class VariableUtility {
     public static isVariableReference(document: TextDocument, position: Position): boolean {
         let wordRange = document.getWordRangeAtPosition(position);
         let lineRange = document.lineAt(position);
+        return VariableUtility.isVariableReferenceFromLine(wordRange, lineRange);
+    }
+
+    public static isRequestVariableReference(document: TextDocument, position: Position): boolean {
+        let wordRange = document.getWordRangeAtPosition(position, /(\w+)(\.\w+|\[\d+\])+/);
```
Huachao: Update the regex here also?

src/requestVariableHoverProvider.ts
```diff
@@ -0,0 +1,48 @@
...
+                return new Hover(contents, wordRange);
+            }
+        }
+
+        let contents: MarkedString[] = [{ language: 'http', value: `Warning: Request Variable ${fullPath} is not loaded in memory` }];
+        return new Hover(contents, wordRange);
+    }
+
+    private getRequestVariableHoverPath(wordRange: Range, lineRange: TextLine, position: Position) {
+        let index = position.character - 1;
```
Huachao: If we don't consider the case that space is allowed between variable name and `{{...}}`, can we simply remove the `for` loop

src/requestVariableCacheValueProcessor.ts
cbrevik: Well, `response.headers` seem to be an object. Though the property-names are actual header names like `Content-Type`. So one way to support it would be to allow using string values like `variable.response.headers["Content-Type"]`. Requires changing a few regexes though.

src/requestVariableCacheValueProcessor.ts
cbrevik: Sure!

src/requestVariableCacheValueProcessor.ts
cbrevik: Well, I looked into this, and couldn't quite decide how to solve it. If there are any suggestions here they are welcome.

The issue, as I saw it, is that not the whole object-structure is a JSON-string. Sure we could allow JSON path for the `body` of `response`, but what about the rest of the object? Should JSON path start from the `body`? 

`variableName.response.body$.users[:1].id`?

I couldn't decide on a good approach, so I just did a simple naive manual resolving of values.

src/variableDiagnosticsProvider.ts
cbrevik: Well, `code` is a required field, but I guess I could set it to `0`?

src/variableDiagnosticsProvider.ts
cbrevik: Sure I'll look into it. This was just the recommended pattern when I looked at tutorials from Microsoft.

src/events/requestVariableEvent.ts
cbrevik: I'll export it, we need some type to use with `EventEmitter`. One idea which I didn't get time to complete was checking a specific document rather than "all" when firing the event. In that case the cacheKey is useful to send with the event.

src/variableDiagnosticsProvider.ts
cbrevik: Sure. I'll rename it.

src/extension.ts
```diff
@@ -59,7 +59,13 @@ export async function activate(context: ExtensionContext) {
     context.subscriptions.push(languages.registerDefinitionProvider('http', new CustomVariableDefinitionProvider()));
     context.subscriptions.push(languages.registerReferenceProvider('http', new CustomVariableReferenceProvider()));
     context.subscriptions.push(languages.registerDocumentSymbolProvider('http', new HttpDocumentSymbolProvider()));
-    new VariableDiagnosticsProvider().activate(context.subscriptions);
+
+    const diagnosticsProviders = new VariableDiagnosticsProvider();
+    workspace.onDidOpenTextDocument(diagnosticsProviders.checkVariables, this, context.subscriptions);
```
Huachao: It seems the `this` target is not right, we should use `diagnosticsProviders` instead.

src/requestVariableCacheValueProcessor.ts
Huachao: Can we use syntax like `variable.response.header.xxx`, (xxx represents the actual header name like `Content-Type`, `X-RateLimit-Limit`, `ETag`)

src/variableDiagnosticsProvider.ts
```diff
@@ -51,23 +45,23 @@ export class VariableDiagnosticsProvider {
...
-                    if (v.variableName === ea.name) {
+                    if (v.variableName === name) {
                         diagnostics.push({
                             severity: DiagnosticSeverity.Error,
                             range: new Range(new Position(v.lineNumber, v.startIndex), new Position(v.lineNumber, v.endIndex)),
                             message: `${v.variableName} is not loaded in memory`,
-                            source: 'ex',
-                            code: "10",
+                            source: 'http',
+                            code: 0,
```
Huachao: do we need this line, what is `code` used for?

src/events/requestVariableEvent.ts
Huachao: @cbrevik that's OK, we can improve it later. Currently it's OK for me.

src/requestVariableCacheValueProcessor.ts
Huachao: @cbrevik my suggestion is that make the syntax like `variableName.response.body.$.users...`
That means the whole syntax is likes:
```
variableName.(response|request).(header|body).(header name|JSON Path)
```

src/constants.ts
```diff
@@ -60,4 +60,6 @@ export const AzureClouds = {
 
 export const CommentIdentifiersRegex: RegExp = /^\s*(#|\/{2})/;
 
-export const VariableDefinitionRegex: RegExp = /^\s*@([^\s=]+)\s*=\s*(.+)\s*$/;
\ No newline at end of file
+export const VariableDefinitionRegex: RegExp = /^\s*@([^\s=]+)\s*=\s*(.+)\s*$/;
+
+export const RequestVariableDefinitionRegex: RegExp = /^\s*#{1,}\s+@name\s+(\S+)\s*$/m;
```
Huachao: I think we should also support line starts with `//`, since double slash is also support as line comment indicator.

src/variableDiagnosticsProvider.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export class VariableDiagnosticsProvider {
+    private httpDiagnosticCollection: DiagnosticCollection;
+
+    constructor() {
+        this.httpDiagnosticCollection = languages.createDiagnosticCollection();
+
+        this.checkVariablesInAllTextDocuments();
+
+        OnRequestVariableEvent(this.checkVariablesInAllTextDocuments);
```
Huachao: Sorry, you should still use __arrow function__ syntax, otherwise would cause an incorrect `this` binding when callback `checkVariablesInAllTextDocuments` called.

Huachao:
@cbrevik I uploaded a new commit which adds the functionality to highlight the request variable in `http` file.

src/variableDiagnosticsProvider.ts
cbrevik: I think this is some old logic I forgot to remove, we don't need the split for either `.` or `[`. Fixing.

src/variableDiagnosticsProvider.ts
cbrevik: The `code` parameter is required, so I have to set it to something

src/extension.ts
cbrevik: Will fix

src/variableDiagnosticsProvider.ts
cbrevik: Well, `\S` allows `}`. Which makes for example `GET {{baseUrl}}/users/{{variableName.response.body.userId}}` difficult to match (it will match the whole thing except `GET`.

I think we should be a bit strict, possibly agree on a set of legal characters. 

We could of course write even more complex and difficult to understand regexes, but is there a lot of value in that?

src/constants.ts
cbrevik: I'm unsure about this change now. `\S` seems a bit too broad, and will allow characters that make the regex even more complex.

src/variableDiagnosticsProvider.ts
cbrevik: For example we probably have to allow `-` because of the header-names. But allowing any `$`, `[` or `{` etc outside of property/JSON-path lookups will make the regexes hard to use. I think we should land a consensus here before I do any more changes to regexes.

cbrevik:
My regex comment re-iterated;
`\S` allows `}`. Which makes for example `GET {{baseUrl}}/users/{{variableName.response.body.userId}}` difficult to match (it will match the whole thing except `GET`.)

I think we should be a bit strict, possibly agree on a set of legal characters.

We could of course write even more complex and difficult to understand regexes, but is there a lot of value in that?

For example we probably have to allow `-` because of the header-names. But allowing any `$`, `[` or `{` etc outside of property/JSON-path lookups will make the regexes hard to use. I think we should land a consensus here before I do any more changes to regexes.

Huachao:
@cbrevik I agree with your opinion towards the regex, I think the `\S` is a bit broader, let's restrict the special characters to `-`, `$`, `[`, `]`, `{`, `}` and `_`. Do you think it's OK?

src/variableDiagnosticsProvider.ts
Huachao: @cbrevik I agree with your opinion towards the regex, I think the `\S` is a bit broader, let's restrict the special characters to `-`, `$`, `[`, `]`, `{`, `}` and `_`. Do you think it's OK?

cbrevik: Sounds okay. I think most of this will play out in `RequestVariableCacheValueProcessor`, the specific line we're commenting now is diagnostic for **all** variables (not just request variables). I'll update and ping you back.

Huachao: @cbrevik yes, the regex to match the request variable should be broader, and the details reference path(including`.(request|response).(header|body).(path|name)`) should be parsed and extracted in the `RequestVariableCacheValueProcessor`. And the parsing result can be used both for replacing the request with actual value, and diagnostic help(tell the user the which part of path is incorrect, like first part should be request or response, second part can only ref header and body, third part should be a exist header name or a correct JSON path of JSON body).

cbrevik: So we should only allow `headers` or `body` as the second part as you say?

I suppose I could narrow it down in completion and diagnostic.

Just for reference, this is the current auto-completion of `response`:
image

`request`:
image

Huachao: @cbrevik personally I think we can narrow down to _body_ and _header_ only, I think it still would be easier to allow other options later. We just need to modify the regex and the `RequestVariableCacheValueProcessor`.

cbrevik:
I'm working on locking down the completion/diagnostic allowed paths for variables to `(variableName).(request|response).(headers|body).(jsonpath|name)` as discussed. As well as supporting `headers`. That is not too difficult.

What I'm having problem with is thinking out how we should support JSONPath. The library I am looking at is; https://github.com/dchester/jsonpath (please comment if anyone has a better suggestion).

Every query for JSONPath always returns an array. So how should we decide what to select for? 
We could just say, just take `result[0]` if it exists. But what if someone actually wants to retrieve the whole array? 

I'm not sure how to delimit that outside of the JSONPath. Should we invent syntax for that as well? Seems a bit silly maybe?

At the moment I'm thinking maybe JSONPath should be considered for another PR, and my manual retrieval (`JSON.parse(body)`) should stay as it is. What do you think @Huachao?

Huachao:
@cbrevik, sorry for the late reply, since I am approaching China Spring Festival. Personally I think we can use the jsonpath package you've mentioned, and it's OK for me to only allow user to fetch the first item in array. If the first item is of string type, simply returns, otherwise JSON.stringify() the first item. I think it should be enough for most of our user cases, we can document this behavior in README.

cbrevik:
Alright @Huachao, implementation is updated to allow JSONPath + the current format for request variables:`(variableName).(request|response).(headers|body).(jsonpath|name)`.

It's was a bit tricky providing proper hover- & completion-support with JSONPath. So I updated the implementation to retrieve the value from the whole path when hovering.

src/requestVariableCacheValueProcessor.ts
```diff
@@ -1,114 +1,83 @@
...
         }
+    }
 
-        return body;
+    public static resolveHttpBody(body: any, path: string) {
+        try {
+            const result = jp.query(body, path);
+            return result ? result[0] : null;
+        } catch (err) {
+            console.warn(`The JSONPath query failed, is ${path} correctly formatted?`);
```
Huachao: Notify this to user with window?

src/requestVariableCacheValueProcessor.ts
```diff
@@ -1,114 +1,83 @@
...
+        } else {
+            return undefined;
         }
+    }
 
-        return body;
+    public static resolveHttpBody(body: any, path: string) {
+        try {
+            const result = jp.query(body, path);
+            return result ? result[0] : null;
```
Huachao: do we also notify user a warning that the json path query result is an array(more than one item) which is not supported.

src/requestVariableCacheValueProcessor.ts
cbrevik: I did that initially, but `requestVariableCacheValueProcessor` is just by both hover & completion. Which means (on syntax errors), that it will pop up an error/warning each time you hover over the JSONPath.

src/requestVariableCacheValueProcessor.ts
cbrevik: Maybe it should? Or at least note it in docs. But same answer as above, it would spam the user on repeated hovers.

src/requestVariableCacheValueProcessor.ts
Huachao: @cbrevik OK, just keep untouch, and I will document these in README.

src/requestVariableCacheValueProcessor.ts
```diff
@@ -1,114 +1,83 @@
...
 import { RequestVariableCacheValue } from './models/requestVariableCacheValue';
 import { HttpResponse } from './models/httpResponse';
+import { HttpRequest } from "./models/httpRequest";
+
+const jp = require('jsonpath');
 
-const validPathRegex = new RegExp(/(\w+)(\.\w+|\[\d+\])*/);
-const partRegex = new RegExp(/(\w+)(\[\d+\])*/);
-const arrayIndexRegex = new RegExp(/\[(\d+)\]/);
+const requestVariablePathRegex: RegExp = /(\w+)(?:\.)?(response|request)?(?:\.)?(body|headers)?(?:\.)?(.*)?/;
```
Huachao: why add some `?` after some groups, they can be omitted?

cbrevik: It's more about not needing a complete path in order to resolve a value. `var.response` is a valid path, `var.response.body` also. In other words, you don't need the complete `var.response.body.$.id` to resolve a value.

I'm not sure what use-cases it is, but it doesn't hurt to support it.

cbrevik: In this way you can hover over `{{var.response.body}}` to see what `body` contains.

Huachao: @cbrevik in this case, we allow `body` on when `response` or `request` is provided, we can't write statements like `variable.body.$.id`, it should be `variable.response.body.$.id`.
So the regex maybe should be like this:
```js
/(\w+)(?:\.(request|response)(?:\.(body|headers)(?:\.(.*))?)?)?/
```

src/requestVariableCacheValueProcessor.ts
```diff
@@ -1,114 +1,83 @@
...
+                if (!http) {
+                    return undefined;
+                }
+
+                if (!httpPart) {
+                    return http;
+                }
+
+                return RequestVariableCacheValueProcessor.resolveHttpPart(http, httpPart, rest);
             default:
```
Huachao: when will this default branch hit?

src/requestVariableCacheValueProcessor.ts
Huachao: I think we should check the type of result[0] here, if the result is `string`, just simply return, otherwise, `JSON.stringify(value)` it. The difference here is that `stringify` a raw string is not equivalent to its original one.  
image

src/requestVariableCacheValueProcessor.ts
cbrevik: Probably never; but as you've noted above, regex validation is not infallible. So I'd rather keep.

src/requestVariableCacheValueProcessor.ts
cbrevik: Ah sure, I'll look into that.

src/requestVariableCacheValueProcessor.ts
cbrevik: Good point, I'll update this

src/requestVariableCacheValueProcessor.ts
Huachao: OK

src/requestVariableCacheValueProcessor.ts
cbrevik: I instead moved stringified outside of `requestVariableCacheValueProcessor`, because `requestVariableCompletionItemProvider` needs the actual (non-string) value to provide completion items of objects.

src/requestVariableCacheValueProcessor.ts
cbrevik: Updated

src/variableDiagnosticsProvider.ts
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    public async checkVariables(document: TextDocument) {
+        if (document.languageId !== 'http') {
+            return;
+        }
+
+        let diagnostics: Diagnostic[] = [];
+
+        let vars = this.findVariables(document);
```
Huachao: How about make the `findVariables` return `Map<string, Variable>`, so that the following logic be more simple and  achieve better performance.

```js
const varNames = [...new Set(vars.keys())]
...
if (!exists) {
  const v = vars.get(name);
diagnostics.push...
}
```

src/variableDiagnosticsProvider.ts
```diff
@@ -0,0 +1,101 @@
...
+            }
+        });
+
+        this.httpDiagnosticCollection.set(document.uri, diagnostics);
+    }
+
+    private findVariables(document: TextDocument): Variable[] {
+        let vars: Variable[] = [];
+        let lines = document.getText().split(/\r?\n/g);
+        let pattern = /\{\{(\w+)(\..*?)*\}\}/;
```
Huachao: add a `g` options for the pattern seems can finally find all the expected variables

And the logic in the while statement can be easier(no need to manually adjust the current index)
```js
while (match = pattern.exec(line)) {
    let variablePath = match[0];
    let variableName = match[1];
    vars.push(new Variable(
        variableName,
        match.index,
        match.index + variablePath.length,
        i
    ));
}
```

src/models/variableType.ts
```diff
@@ -0,0 +1,8 @@
+"use strict";
+
+export enum VariableType {
+    Custom,
```
Huachao: How about rename `Custom` to `File`

src/variableProcessor.ts
```diff
@@ -366,4 +398,70 @@ export class VariableProcessor {
...
+
+    public static async checkVariableDefinitionExists(document: TextDocument, variableNames: string[]): Promise<{ name: string, exists: boolean }[]> {
+        let definitions = await VariableProcessor.getVariableDefinitionsInFile(document);
+
+        return variableNames.map((name) =>
+            ({ name, exists: definitions.has(name) })
+        );
+    }
+
+    public static async getVariableDefinitionsInFile(document: TextDocument): Promise<Map<string, VariableType[]>> {
```
Huachao: Make modifier to __private__

src/requestVariableCompletionItemProvider.ts
```diff
@@ -0,0 +1,97 @@
...
+import { CompletionItemProvider, TextDocument, Position, CancellationToken, CompletionItem, CompletionItemKind, Range, TextLine } from 'vscode';
+
+import { RequestVariableCacheValueProcessor } from "./requestVariableCacheValueProcessor";
+import { VariableProcessor } from "./variableProcessor";
+import { VariableUtility } from "./variableUtility";
+
+const firstPartRegex: RegExp = /^(\w+)\.$/;
+const secondPartRegex: RegExp = /^(\w+)\.(request|response)\.$/;
+
+export class RequestVariableCompletionItemProvider implements CompletionItemProvider {
```
Huachao: Do we need to only allow these completion items when the first part or so called _variable name_ is a request variable?

src/variableDiagnosticsProvider.ts
cbrevik: Making a Map<string, Variable> here would disallow re-use of the same variable across lines in the document. Could make it a Map<string, Variable[]> and re-update the array-value I suppose, but I'm not sure that is more performant.

Huachao: @cbrevik sorry, it should be `Map<string, Variable[]>`. The reason I think more performant is that, in your case, the `vars` will be iterated all over for each non exist variable name to find the matched ones:
```js
existArray.forEach(({ name, exists }) => {
    if (!exists) {
        vars.forEach((v) => {
            if (v.variableName === name) {
                diagnostics.push({
                    severity: DiagnosticSeverity.Error,
                    range: new Range(new Position(v.lineNumber, v.startIndex), new Position(v.lineNumber, v.endIndex)),
                    message: `${v.variableName} is not loaded in memory`,
                    source: 'http',
                    code: 0,
                });
            }
        });
    }
});
```
While what I think we can improve is that we only need to iterate the array which corresponds to the variable name:
```js
existArray.forEach(({ name, exists }) => {
    if (!exists) {
        vars.get(name).forEach((v) => {
            diagnostics.push({
                severity: DiagnosticSeverity.Error,
                range: new Range(new Position(v.lineNumber, v.startIndex), new Position(v.lineNumber, v.endIndex)),
                message: `${v.variableName} is not loaded in memory`,
                source: 'http',
                code: 0,
            });
        });
    }
});
```

cbrevik: Alright, I was thinking it was possibly not more performant because of more time complexity in `findVariables`, with the need to lookup if variable already exists in Map, and then push a new variable into the Array for that Map-key. So whatever gains we have from lookup afterwards is lost in the complexity of building the Map in the first case.

Although, I haven't really done the math, and I don't think it is that complex to really matter either way. Your way looks better as well, so I'll refactor according to your suggestion.

src/variableProcessor.ts
cbrevik: Sure, was thinking this is a utility-method which can be used in the future. But it can be made public again then. I'll update 👍

src/requestVariableCompletionItemProvider.ts
cbrevik: You mean checking if the request variable exists at all in memory? I'm not sure that is developer friendly, personally I like having completion on "known" valid syntax, even if it won't run.

Is there a good reason to do that?

src/variableDiagnosticsProvider.ts
cbrevik: This has been refactored. 👍

src/requestVariableCompletionItemProvider.ts
Huachao: @cbrevik  I don't express myself clearly, I just mean this completion item hint only make sense for request variable, not for other types of variables. So it's not necessary for request variables loaded in memory, just exist or already declared in file is enough.

cbrevik: I was thinking maybe the check at the top was enough;
```typescript
if (!VariableUtility.isPartialRequestVariableReference(document, position)) {
    return [];
}
```

Which basically checks if the path is using a dot, which would by definition would be a request variable. (Since only that type uses dotted values).

But it does not check if the variable name is declared in the file as a request variable, is that what you're thinking should also be checked?

cbrevik: I've added a check for if the variable name is defined in the file as a request variable (and not necessarily in memory)

cbrevik:
Think I've adressed the comments now :+1:

src/requestVariableCompletionItemProvider.ts
```diff
@@ -67,6 +74,19 @@ export class RequestVariableCompletionItemProvider implements CompletionItemProv
         return completionItems;
     }
 
+    private checkIfRequestVariableDefined(document: TextDocument, variableName: string) {
+        const text = document.getText();
+        const regex = new RegExp(Constants.RequestVariableDefinitionRegex, 'mg');
```
Huachao: Why not directly construct a regex with given variableName, so that if we can found one in `text` means `True`, otherwise `False`.

src/requestVariableCompletionItemProvider.ts
```diff
@@ -0,0 +1,117 @@
...
+                        : fullPath;
+
+                const valueAtPath = RequestVariableCacheValueProcessor.getValueAtPath(variableValue, fullPath);
+                if (valueAtPath) {
+                    let props = Object.getOwnPropertyNames(valueAtPath);
+
+                    completionItems = props.map(p => {
+                        let item = new CompletionItem(p);
+                        item.detail = `(property) ${p}`;
+                        item.documentation = p;
```
Huachao: How about make the value of `documentation` to the actual value of `p`? User can better know what's the actual value will be inserted into the request.

src/requestVariableCompletionItemProvider.ts
```diff
@@ -0,0 +1,117 @@
...
+                        ? fullPath.substring(0, fullPath.length - 1)
+                        : fullPath;
+
+                const valueAtPath = RequestVariableCacheValueProcessor.getValueAtPath(variableValue, fullPath);
+                if (valueAtPath) {
+                    let props = Object.getOwnPropertyNames(valueAtPath);
+
+                    completionItems = props.map(p => {
+                        let item = new CompletionItem(p);
+                        item.detail = `(property) ${p}`;
```
Huachao: Why add the _(property)_ prefix

src/requestVariableCompletionItemProvider.ts
```diff
@@ -0,0 +1,117 @@
...
+            if (regex.test(fullPath)) {
+                // Remove last dot if present
+                fullPath =
+                    fullPath.charAt(fullPath.length - 1) === "."
+                        ? fullPath.substring(0, fullPath.length - 1)
+                        : fullPath;
+
+                const valueAtPath = RequestVariableCacheValueProcessor.getValueAtPath(variableValue, fullPath);
+                if (valueAtPath) {
+                    let props = Object.getOwnPropertyNames(valueAtPath);
```
Huachao: Minor: Why not use Object.keys(), since your one will also include not enumerable properties?

cbrevik:
Fixed!

Huachao:
@cbrevik Merged, thanks, really a great patch 👍

cbrevik:
Awesome, thanks for the help landing this @Huachao! It's going to be fun using this "in production".

PR #142, opened 2017-12-07
Allow hiding status bar item via settings

VFK:
This is such a minor nitpick but it feels a bit distracting for me when i'm editing files that completely unrelated to REST client but that status bar switcher is always present 😄 

This pull request adds `showStatusBarItem` boolean parameter configurable from settings.
Since we can switch environments via command palette anyway we aren't losing in functionality with this new config param set to `false`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -442,7 +442,7 @@
           "scope": "resource",
           "description": "Response preview output option. 'full' for whole response message(status line, headers and body). 'headers' for response headers(as well as status line). 'body' for response body only. 'exchange' for whole HTTP exchange (request and response)"
         },
-        "rest-client.showStatusBarItem": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -441,6 +441,12 @@
           "default": "full",
           "scope": "resource",
           "description": "Response preview output option. 'full' for whole response message(status line, headers and body). 'headers' for response headers(as well as status line). 'body' for response body only. 'exchange' for whole HTTP exchange (request and response)"
+        },
+        "rest-client.showStatusBarItem": {
```
Huachao: minor: how about rename it to showEnvironmentStatusBarItem, since status bar in my extension also includes response duration and size status bars

VFK: @Huachao Sure, sounds good to me

PR #150, opened 2017-12-31
Add built-in support for Azure AD auth

flanakin:
No more copying AAD tokens from the portal or other sources :-)
**`{{$aadToken}}`** &ndash; Get token for home directory of the signed-in user
**`{{$aadToken 00000000-0000-0000-0000-000000000000}}`** &ndash; Get token for directory using tenant ID
**`{{$aadToken contoso.com}}`** &ndash; Get token for directory using domain
**`{{$aadToken new contoso.com}}`** &ndash; Force re-auth and don't reuse tokens across requests
**`{{$aadToken public|cn|de|us|ppe}}`** &ndash; Get token for specific cloud (otherwise implied from URL)
**`F1 > Rest Client: Clear AAD token cache`** &ndash; Clear the in-memory token cache (or restart Code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,7 +370,7 @@ Content-Type: application/json
 Global variables provide a pre-defined set of variables that can be used in any part of the request(Url/Headers/Body) in the format `{{$variableName}}`. Currently, we provide a few dynamic variables which you can use in your requests. The variable names are _case-sensitive_.
 * `{{$aadToken [new] [public|cn|de|us|ppe] [<domain|tenantId>]}}`: Add an Azure Active Directory token based on the following options (must be specified in order):
 
-  `new`: Optional. Specify `new` to force re-authentication and get a new token for the specified directory. Default: Reuse non-expired token for the specified directory from an in-memory cache. (Restart Code to clear the cache.)
+  `new`: Optional. Specify `new` to force re-authentication and get a new token for the specified directory. Default: Reuse non-expired token for the specified directory from an in-memory cache. (Use `F1 > Rest Client: Clear Azure AD token cache` or restart Code to clear the cache.)
 
   `public|cn|de|us|ppe`: Optional. Specify top-level domain (TLD) to get a token for the specified government cloud, `public` for the public cloud, or `ppe` for internal testing. Default: TLD of the REST endpoint; `public` if not valid.
 
@@ -442,7 +442,7 @@ GET https://management.azure.com/subscriptions
 Authorization: {{$aadToken new contoso.com}}
 ```
 
-_**NOTE:** REST Client uses an in-memory token cache that clears when Code is restarted._
+_**NOTE:** REST Client uses an in-memory token cache that clears when Code is restarted. You can clear the token cache manually by using `F1 > Rest Client: Clear Azure AD token cache`._
 
 _**NOTE:** `new` can be used with any other options, as long as it's specified first. Order is important for all options._
 
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ REST Client allows you to send HTTP request and view the response in Visual Stud
     - Support both __environment__ and __file__ custom variables
     - Auto completion and hover support for both environment and file custom variables
     - Go to definition and find all references support _ONLY_ for file custom variables
-    - Provide system dynamic variables `{{$guid}}`, `{{$randomInt min max}}`, `{{$timestamp}}`, and `{{$aadToken [tenantId]}}`
+    - Provide system dynamic variables `{{$guid}}`, `{{$randomInt min max}}`, `{{$timestamp}}`, and `{{$aadToken [domain|tenantId]}}`
     - Easily create/update/delete environments and custom variables in setting file
     - Support environment switch
     - Support shared environment to provide variables that available in all environments
@@ -368,8 +368,8 @@ Content-Type: application/json
 Global variables provide a pre-defined set of variables that can be used in any part of the request(Url/Headers/Body) in the format `{{$variableName}}`. Currently, we provide a few dynamic variables which you can use in your requests. The variable names are _case-sensitive_.
 * `{{$guid}}`: Add a RFC 4122 v4 UUID
 * `{{$randomInt min max}}`: Returns a random integer between min (included) and max (excluded)
-* `{{$aadToken [tenantId]}}`: Add `Authorization: {{$aadToken}}` to attach an Azure Active Directory token to the request
-  *tenantId*: Optional. Unique identifier (GUID) for the directory. Default is account's home directory.
+* `{{$aadToken [directory]}}`: Add `Authorization: {{$aadToken}}` to attach an Azure Active Directory token to the request
+  `directory`: Optional. Domain or GUID for the directory to sign in to. Default is account's home directory.
 * `{{$timestamp}}`: Add UTC timestamp of now. You can even specify any date time based on current time in the format `{{$timestamp number option}}`, e.g., to represent 3 hours ago, simply `{{$timestamp -3 h}}`; to represent the day after tomorrow, simply `{{$timestamp 2 d}}`. The option string you can specify in timestamp are:
 
 Option | Description
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ REST Client allows you to send HTTP request and view the response in Visual Stud
     - Support both __environment__ and __file__ custom variables
     - Auto completion and hover support for both environment and file custom variables
     - Go to definition and find all references support _ONLY_ for file custom variables
-    - Provide system dynamic variables `{{$guid}}`, `{{$randomInt min max}}`, `{{$timestamp}}`, and `{{$aadToken}}`
+    - Provide system dynamic variables `{{$guid}}`, `{{$randomInt min max}}`, `{{$timestamp}}`, and `{{$aadToken [tenantId]}}`
     - Easily create/update/delete environments and custom variables in setting file
     - Support environment switch
     - Support shared environment to provide variables that available in all environments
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -366,6 +368,14 @@ Content-Type: application/json
 
 ### Global Variables
 Global variables provide a pre-defined set of variables that can be used in any part of the request(Url/Headers/Body) in the format `{{$variableName}}`. Currently, we provide a few dynamic variables which you can use in your requests. The variable names are _case-sensitive_.
+* `{{$aadToken [new|clear] [public|cn|de|us|ppe] [<domain|tenantId>]}}`: Add an Azure Active Directory token based on the following options (must be specified in order):
+
```
Huachao: I don't like adding the `clear` code action directly inside the document, which makes ambiguous with using a variable value or url. So how about add a command directly in vscode, or simply remove this option.

src/variableProcessor.ts
```diff
@@ -40,9 +79,130 @@ export class VariableProcessor {
...
+        if (cloud === null) {
+            cloud = targetApp.substring(targetApp.lastIndexOf(".") + 1, targetApp.length - 1);
+        }
+
+        // parse input options -- [new|clear] [public|cn|de|us|ppe] [<domain|tenantId>]
+        let tenantId = "common";
+        let forceNewToken = false;
+        const groups = new RegExp(aadRegexPattern).exec(url);
+        if (groups) {
+            forceNewToken = (groups.length > 2 && groups[2] === Constants.AzureActiveDirectoryForceNewOption);
```
Huachao: no need to check groups array length here, it will safely return undefined if out of the array range

src/variableProcessor.ts
```diff
@@ -1,11 +1,50 @@
...
+}
+
+// see TokenResponse at https://github.com/AzureAD/azure-activedirectory-library-for-nodejs/blob/dev/lib/adal.d.ts
+interface SignInToken {
+    tokenType: string;
+    expiresIn: number;
+    expiresOn: Date | string;
+    resource: string;
+    accessToken: string;
+    refreshToken?: string;
```
Huachao: Would you like to support automatically refreshing access token when expired?

README.md
flanakin: Good idea. A command sounds like the right approach for this. I'll remove it for now and re-add it as a command later.

README.md
```diff
@@ -398,6 +408,72 @@ X-Request-Id: {{token}}
...
+Do not reuse older token -- force re-authentication for current directory (e.g. switch account):
+
+```http
+GET https://management.azure.com/subscriptions
+    ?api-version=2017-08-01
+Authorization: {{$aadToken new contoso.com}}
+```
+
+Clear in-memory tocken cache -- force re-authentication for all directories:
```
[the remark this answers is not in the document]
flanakin: Removed this as part of removing the clear option

README.md
```diff
@@ -398,6 +408,72 @@ X-Request-Id: {{token}}
...
+
+```http
+GET https://management.azure.com/subscriptions
+    ?api-version=2017-08-01
+Authorization: {{$aadToken clear contoso.com}}
+```
+
+_**NOTE:** REST Client uses an in-memory token cache that clears when Code is restarted.
```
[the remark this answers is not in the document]
flanakin: Good catch!

src/constants.ts
```diff
@@ -22,6 +22,35 @@ export const GuidVariableDescription = "Add a RFC 4122 v4 UUID";
 export const RandomInt = "$randomInt";
 export const RandomIntDescription = "Returns a random integer between min (included) and max (excluded)";
 
+export const AzureActiveDirectoryVariableName = "$aadToken";
+export const AzureActiveDirectoryDescription = "Prompts to sign in to Azure AD and adds the token to the request";
+export const AzureActiveDirectoryClientId = "aebc6443-996d-45c2-90f0-388ff96faa56";  // VS Code client id
```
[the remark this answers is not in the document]
flanakin: I'll add comments to make this clearer in the code...

```javascript
/**
 * NOTE: The client id represents an AAD app people sign in to. The client id is sent to AAD to indicate what app
 * is requesting a token for the user. When the user signs in, AAD shows the name of the app to confirm the user is
 * authorizing the right app to act on their behalf. We're using Visual Studio Code's client id since that is the
 * overarching app people will think of when they are signing in.
 */
```

src/variableProcessor.ts
```diff
@@ -1,11 +1,50 @@
 'use strict';
 
-import { window } from 'vscode';
+import { commands, Uri, window } from 'vscode';
 import { EnvironmentController } from './controllers/environmentController';
 import * as Constants from './constants';
 import { Func } from './common/delegates';
 import * as moment from 'moment';
 const uuid = require('node-uuid');
+const adal = require('adal-node');
```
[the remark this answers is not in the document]
flanakin: Want me to do the same for node-uuid while I'm in there?

src/variableProcessor.ts
```diff
@@ -1,11 +1,50 @@
...
 import * as moment from 'moment';
 const uuid = require('node-uuid');
+const adal = require('adal-node');
+const copyPaste = require('copy-paste');
+
+const aadRegexPattern = `\\{\\{\\s*\\${Constants.AzureActiveDirectoryVariableName}(\\s+(${Constants.AzureActiveDirectoryForceNewOption}|${Constants.AzureActiveDirectoryClearCacheOption}))?(\\s+(ppe|public|cn|de|us))?(\\s+([^\\.]+\\.[^\\}\\s]+|[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}))?\\s*\\}\\}`;
+const aadTokenCache = {};
+
+// see UserCodeInfo at https://github.com/AzureAD/azure-activedirectory-library-for-nodejs/blob/dev/lib/adal.d.ts
+interface SignInCode {
```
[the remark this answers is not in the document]
flanakin: For some reason, I was getting an error when I tried that. I gave up debugging and just went the simple route. I saw other Microsoft repos doing this. Regardless, seems like I was able to update it. Who knows :-?

src/variableProcessor.ts
flanakin: Yeah, I put it in my backlog. It'd save a re-auth scenario, so it'd definitely be useful. Not a blocker for basic functionality, tho.

src/variableProcessor.ts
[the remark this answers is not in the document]
Huachao: I am afraid not. In fact both of these syntaxes will be compiled to require statement in NodeJs enviroment (depends on the tsconfig.json). Here I want to update this module import syntax is that declaring in TypeScript way will have benefits from editor to get intellisense when using this module. And the root cause is that, this module has its typing file in its code already and specified in its [package.json](https://github.com/AzureAD/azure-activedirectory-library-for-nodejs/blob/dev/package.json#L24)

src/variableProcessor.ts
```diff
@@ -40,9 +79,130 @@ export class VariableProcessor {
             }
         }
 
+        // promise-based variables using URL-inspection must be last
```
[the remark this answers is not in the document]
flanakin: I started with `$aadToken` in `getGlobalVariables()`, but it has 2 unique traits compared to other global vars:
1. I need the request URL to tell AAD who the token will be sent to
2. Sign-in uses callbacks

Adding the URL isn't a big deal, but making `getGlobalVariables()` callback-/promise-aware quickly became unnecessarily complicated, so I decided to keep it simple and let this be a unique case. We should revisit this if/when there are other promise-based vars.

Next, I moved this code after global vars are evaluated, which worked great until I added a custom variable in the `aadToken.http` test file (e.g. `GET {{arm}}/subscriptions`). This caused a problem because AAD won't give me a token for `{{arm}}` &ndash; we need to evaluate the custom var _before_ we parse the URL. That's what I meant by "URL-inspection".

Let me know if you think there's a better way to convey that than "URL-inspection". I struggled with it for a while, but moved on hoping I'd come up with a better way to describe it. Clearly, I haven't :-(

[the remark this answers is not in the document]
Huachao: Oh, understand, totally agree with what you've done here. It makes sense.

src/variableProcessor.ts
flanakin: Yeah, I saw there was a placeholder for every regex lookup, but I guess I was opting for safer code. Given we don't have a cross-browser situation here, you're right that this isn't necessary. I'll remove it.

src/variableProcessor.ts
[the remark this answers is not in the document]
Huachao: Maybe I am wrong, I guess the error is a compiler error from TypeScript. And I can compile successfully when I replace the `require` syntax to `import` syntax, and update the reference places to `adal.UserCodeInfo` and `adal.TokenResponse` respectively.

And you can show me the error details if you wish. 😄

src/variableProcessor.ts
Huachao: @flanakin since vscode extension run with Electron, which uses Chrome V8 internally. I think it's safe. 😄

src/variableProcessor.ts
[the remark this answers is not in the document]
flanakin: Changed the comment. Hopefully this helps a little: `// any vars that make decisions on the URL (request var) must be last so all vars have been evaluated`

flanakin:
Thanks for all the feedback! Let me know if you notice anything else that needs to go in for this PR. (And thanks for building such a great extension!)

src/variableProcessor.ts
```diff
@@ -4,48 +4,14 @@ import { commands, Uri, window } from 'vscode';
 import { EnvironmentController } from './controllers/environmentController';
 import * as Constants from './constants';
 import { Func } from './common/delegates';
+import * as adal from 'adal-node';
+import * as copyPaste from 'copy-paste';
```
Huachao: keep `copy-paste` and `node-uuid` modules uses the require syntax, since we don't add their typing files explicitly in `package.json`, so no intellisense will have from Vscode.

I encourage to use typescript module import syntax for `adal-node` is that, this package has typings file in its package directly.

src/variableProcessor.ts
```diff
@@ -117,26 +89,14 @@ export class VariableProcessor {
...
-                for (let key in aadTokenCache) {
-                    delete aadTokenCache[key];
-                }
-
-                // if only clearing, return
-                if (url === `{{${Constants.AzureActiveDirectoryVariableName} ${Constants.AzureActiveDirectoryClearCacheOption}}}`) {
-                    return null;
-                }
-            }
+            forceNewToken = (groups[2] === Constants.AzureActiveDirectoryForceNewOption);
```
Huachao: remove useless parenthese

src/variableProcessor.ts
```diff
@@ -40,9 +45,124 @@ export class VariableProcessor {
...
+        // any vars that make decisions on the URL (request var) must be last so all vars have been evaluated
+        let aadRegex = new RegExp(aadRegexPattern, 'g');
+        if (aadRegex.test(request)) {
+            request = request.replace(aadRegex, await VariableProcessor.getAadToken(request));
+        }
+
         return request;
     }
 
+    public static async clearAadTokenCache() {
```
Huachao: If we want this method to be invoked by VS Code command, we need to update the __extension.ts__ and __package.json__ too.

Huachao:
@flanakin I have three comments, and other parts LGTM. And thanks for your contribution to introduce AAD functionality into this extension

src/variableProcessor.ts
flanakin: Well, I kept this around because I didn't want to add the command right now, but I didn't realize how easy it was going to be :-P Thanks for the pointer!

Huachao:
@flanakin Merged, thanks. Really fantastic patch, I will publish this great patch in next release.

PR #151, opened 2018-01-06
Refresh expired tokens

flanakin:


src/variableProcessor.ts
```diff
@@ -107,51 +107,88 @@ export class VariableProcessor {
         const authContext: any = new adal.AuthenticationContext(signInUrl);
```
Huachao: seems not necessary to set the type of `authContext` to `any`

PR #153, opened 2018-01-08
Pick $aadToken directory when not specified

flanakin:


src/variableProcessor.ts
```diff
@@ -184,6 +187,63 @@ export class VariableProcessor {
...
+                                    let displayName = element.displayName;
+                                    if (displayName === Constants.AzureActiveDirectoryDefaultDisplayName) {
+                                        const separator = domain.indexOf(".");
+                                        displayName = `${separator > 0 ? domain.substring(0, separator) : domain} (${displayName})`;
+                                    }
+                                    directories.push({ label: displayName, description: element.tenantId, detail: domain });
+                                });
+
+                                // default to first directory
+                                let result = directories.length && directories[0];
```
Huachao: seems only `let result = directories[0]` is safe enough.

Huachao:
@flanakin merged, thanks. Another great patch 👍

PR #154, opened 2018-01-13
Try to pick the best domain for $aadToken directories

flanakin:


src/variableProcessor.ts
```diff
@@ -198,16 +198,62 @@ export class VariableProcessor {
...
+                                            }
+
+                                            // use best match
+                                            for (let m in bestMatches) {
+                                                if (bestMatches[m]) {
+                                                    domain = bestMatches[m];
+                                                    break;
+                                                }
+                                            }
+                                        } catch {
```
Huachao: Can you run successfully locally? I think JS needs an expression for catch statements, refer to https://docs.microsoft.com/en-us/scripting/javascript/reference/try-dot-dot-dot-catch-dot-dot-dot-finally-statement-javascript

src/variableProcessor.ts
```diff
@@ -198,16 +198,62 @@ export class VariableProcessor {
...
+                                    if (count > 1) {
+                                        try {
+                                            // find the best matches
+                                            const displayNameSpaceIndex = displayName.indexOf(" ");
+                                            const displayNameFirstWord = displayNameSpaceIndex > -1
+                                                ? displayName.substring(0, displayNameSpaceIndex)
+                                                : displayName;
+                                            let bestMatches = [];
+                                            for (let d in element.domains) {
+                                                const matches = new RegExp(`(^${displayNameFirstWord}\.com$)|(^${displayNameFirstWord}\.[a-z]+(?:\.[a-z]+)?$)|(^${displayNameFirstWord}[a-z]+\.com$)|(^${displayNameFirstWord}[^:]*$)|(^[^:]*${displayNameFirstWord}[^:]*$)`, "gi")
```
Huachao: Put this new Regex logic out of for loop for perf concern

src/variableProcessor.ts
```diff
@@ -198,16 +198,62 @@ export class VariableProcessor {
...
+                                                    .exec(element.domains[d]) || [null, null, null, null, null, null];
+
+                                                // stop looking if the best match is found
+                                                bestMatches[0] = bestMatches[0] || matches[1];
```
Huachao: `bestMatches[0] = matches[1]` is enough

src/variableProcessor.ts
```diff
@@ -198,16 +198,62 @@ export class VariableProcessor {
...
+                                        try {
+                                            // find the best matches
+                                            const displayNameSpaceIndex = displayName.indexOf(" ");
+                                            const displayNameFirstWord = displayNameSpaceIndex > -1
+                                                ? displayName.substring(0, displayNameSpaceIndex)
+                                                : displayName;
+                                            let bestMatches = [];
+                                            for (let d in element.domains) {
+                                                const matches = new RegExp(`(^${displayNameFirstWord}\.com$)|(^${displayNameFirstWord}\.[a-z]+(?:\.[a-z]+)?$)|(^${displayNameFirstWord}[a-z]+\.com$)|(^${displayNameFirstWord}[^:]*$)|(^[^:]*${displayNameFirstWord}[^:]*$)`, "gi")
+                                                    .exec(element.domains[d]) || [null, null, null, null, null, null];
```
Huachao: Simply `Array(6).fill(null)` looks more clearer

src/variableProcessor.ts
```diff
@@ -198,16 +198,62 @@ export class VariableProcessor {
...
+                                                }
+
+                                                bestMatches[1] = bestMatches[1] || matches[2];
+                                                bestMatches[2] = bestMatches[2] || matches[3];
+                                                bestMatches[3] = bestMatches[3] || matches[4];
+                                                bestMatches[4] = bestMatches[4] || matches[5];
+                                            }
+
+                                            // use best match
+                                            for (let m in bestMatches) {
```
Huachao: Seems we can write like `domain = bestMatches.find(m => m) || domain`, just as `FirstOrDefault()` in `C#`

src/variableProcessor.ts
flanakin: Yep, everything works fine. I didn't see anything in the docs that indicted we need something else. Did I miss it? I'll add `catch (e)` just in case.

Huachao: I think JS needs us to write like `catch (e)` instead of `catch`?

src/variableProcessor.ts
```diff
@@ -215,30 +215,28 @@ export class VariableProcessor {
                                                 ? displayName.substring(0, displayNameSpaceIndex)
                                                 : displayName;
                                             let bestMatches = [];
+                                            const bestMatchesRegex = new RegExp(`(^${displayNameFirstWord}\.com$)|(^${displayNameFirstWord}\.[a-z]+(?:\.[a-z]+)?$)|(^${displayNameFirstWord}[a-z]+\.com$)|(^${displayNameFirstWord}[^:]*$)|(^[^:]*${displayNameFirstWord}[^:]*$)`, "gi");
+                                            const bestMatchesRegexGroups = (bestMatchesRegex.source.match(new RegExp(`${displayNameFirstWord}`, "g")) || []).length;
```
Huachao: Do we need this fallback?

src/variableProcessor.ts
```diff
@@ -215,30 +215,28 @@ export class VariableProcessor {
                                                 ? displayName.substring(0, displayNameSpaceIndex)
                                                 : displayName;
                                             let bestMatches = [];
+                                            const bestMatchesRegex = new RegExp(`(^${displayNameFirstWord}\.com$)|(^${displayNameFirstWord}\.[a-z]+(?:\.[a-z]+)?$)|(^${displayNameFirstWord}[a-z]+\.com$)|(^${displayNameFirstWord}[^:]*$)|(^[^:]*${displayNameFirstWord}[^:]*$)`, "gi");
+                                            const bestMatchesRegexGroups = (bestMatchesRegex.source.match(new RegExp(`${displayNameFirstWord}`, "g")) || []).length;
                                             for (let d in element.domains) {
```
Huachao: Personally, I think `for...of` is more appropriate here.

src/variableProcessor.ts
flanakin: This is intended to protect us when the regex is updated to ensure we're always accounting for the correct number of capture groups. I can revert to the explicit 5 checks, but figured this would be more future-proof. I'm fine either way.

Huachao: @flanakin I agree with you that no hard code should exist, I just mean can we remove `|| []` like following:

```js
const bestMatchesRegexGroups = bestMatchesRegex.source.match(new RegExp(`${displayNameFirstWord}`, "g")).length
```

flanakin: Oh, yes! You're right. Since we know it'll match, we'll never have 0. Good catch!

src/variableProcessor.ts
```diff
@@ -216,7 +216,7 @@ export class VariableProcessor {
                                                 : displayName;
                                             let bestMatches = [];
                                             const bestMatchesRegex = new RegExp(`(^${displayNameFirstWord}\.com$)|(^${displayNameFirstWord}\.[a-z]+(?:\.[a-z]+)?$)|(^${displayNameFirstWord}[a-z]+\.com$)|(^${displayNameFirstWord}[^:]*$)|(^[^:]*${displayNameFirstWord}[^:]*$)`, "gi");
-                                            const bestMatchesRegexGroups = (bestMatchesRegex.source.match(new RegExp(`${displayNameFirstWord}`, "g")) || []).length;
+                                            const bestMatchesRegexGroups = (bestMatchesRegex.source.match(new RegExp(`${displayNameFirstWord}`, "g"))).length;
```
Huachao: remove unnecessary ()

Huachao:
@flanakin Merged, thanks 👍

flanakin:
FWIW, I find it annoying clients need to guess the primary domain. I'll try to get ARM to make the primary domain first so we can remove all this extra code.

Huachao:
@flanakin If ARM could provide this information, that will be pretty cool!

PR #160, opened 2018-01-28
Added support for automatic authentication in SharePoint 

s-KaiNet:
### This PR adds support for SharePoint authentication in restclient.    
It's more like POC and question to you, @Huachao if you are OK to merge such kind of PR into your awesome restclient.   

SharePoint has REST API, which allows you to perform various operations with SharePoint objects. However, API requires authentication. Authentication can be complected, because of different authentication schemes in SharePoint as well as different SharePoint versions (online and on-premise). I'm the owner of the module, called [node-sp-auth](https://github.com/s-KaiNet/node-sp-auth) which abstracts different authentication technics and provides seamless api for authentication. In this PR I use `node-sp-auth` for convenient support of SharePoint in your restcilent.    

### How does it work:   
You provide special environmental variable, called `sp-auth`, which points to the file, relative to workspace,  with authentication details. vscode-resclient checks if `sp-auth` variable presented, if so it tries to read authentication details, performs authentication and attaches authentication headers to ongoing request.  Example of env. variable:  
```
"rest-client.environmentVariables": {
        "$shared": {
            "version": "v1"
        },
        "online": {
            "sp-auth": "./online.json"
        },
        "on-prem": {
            "sp-auth": "./onpremise.json"
        },
        "production": {
            "host": "example.com",
            "token": "product token"
        }
    }
```   

### Things to consider:   
 - I had to add two dependencies to your project (probably that's the most controversial part, if you accept new dependencies): [node-sp-auth](https://github.com/s-KaiNet/node-sp-auth) and [node-sp-auth-config](https://github.com/koltyakov/node-sp-auth-config) - config options builder, node module and corresponding CLI  
 - I had to update `httpClient.ts` and include additional call to `await EnvironmentController.getCustomVariables();` which reads current env. file    

Waiting for your feedback :)  and thank you for your great work on restclient.


flanakin:
Doesn't SharePoint use AAD? Have you tried the `$aadToken` support? That hasn't shipped in a release, but is already in the app.

Huachao:
@s-KaiNet First, I'd like to say great work on this patch to support SharePoint authentication in rest client extension. And I have something to discuss with you and @flanakin who introduced Azure Active Directory(AAD) Auth in my extension.

@flanakin I guess __SharePoint online__ maybe work well with AAD, while the on premise version maybe not. However, I am not sure, need to confirm with @s-KaiNet 

@s-KaiNet In your implementation, a HTTP request able to pass the SharePoint auth needs to meet __two__ conditions:

1. Rest client must be an environment with auth config path variable.
2. The auth config file must be in the workspace which requested `http` file exists.

For both conditions, I have following concerns. I think it's a bit implicitly and difficulty to users who want to trigger SP related request at first glance.

* For condition 1:

     * Environment variables are originally intended to replace parts of the content of request file, while here used as for other purpose. 
     * Since you only SP auth indicator into the environment, and no change to individual HTTP requests, which means all the requests running under this environment will all suffer the SP Auth which seems improper.

* For condition 2:
    * All related configs are all defined in specified JSON configuration file, which seems a black box to our extension users. Although SP Auth supports many different kinds of authentication methods, can we still just define them directly in our extension, like what we've done for other supported authentication methods in rest client extension, i.e. __AAD Auth__ by @flanakin, and __Basic Auth__. For example we'd like to put them in each request, can we try to consolidate these different kinds of options into some format like `Authorization: {$spauth auth_method endpoint [auth_user] [auth_password] ...}`, and in our `VariableProcessor.ts` does the work to parse these options and sent them along with your __`node-sp-auth`__.

@s-KaiNet here are just my intuitive thoughts, any discussions and feedbacks are warmly welcomed. And let's work together to make the SP Auth happen in rest client extension. Thanks~

s-KaiNet:
@flanakin yes, I've tried and it indeed works. However, as @Huachao pointed out, it hovers only SharePoint online, there are still on-premise users.    

BTW I have a question to you @flanakin regarding your app registration (Visual Studio Code, client id `aebc6443-996d-45c2-90f0-388ff96faa56`).   
I thought that in order to use the app from other's directory, the app needs to list and allow permissions for a particular resource, SharePoint in my case. However it works perfectly fine for my tenant. It also works fine for MS Graph (`graph.microsoft.com`). Does it mean, that you have added permissions for all possible APIs to those Visual Studio Code app registration? As I've tried to test my own app registration in my tenant without permissions, the error `The client has requested access to a resource which is not listed in the requested permissions in the client's application registration.` is thrown (I used device code flow the same way as you).  My understanding that the app uses app+user permissions, meaning if an app doesn't have sufficient permissions, call will be failed despite user permissions. However your app registration works, that's why I'm curious :)     

@Huachao I agree with all points from both condition `#1` and `#2` and totally understand you. `Authorization: {$spauth...` doesn't work for me, because different auth options set different headers. NTLM options also requires keep alive agent to be set for ongoing `request`. The more I think about it, the more I understand that it simply doesn't fit well with your add-in and the PR should be closed. Not because something wrong with your add-in or my module, but because they are different and solves different problems.    
However I still would like to have support of SharePoint in your awesome addin. For now I can see two possible solutions, will glad to know your personal opinion:  
 1.  Add a possibility to restclient to inject custom authentication data from third party providers. That's the most difficult one. Probably some other third parties would like to add support. Ideally there should be a way to extend restclient without adding new dependencies and without hurting performance of course.   
2. I can publish a new plugin based on your own with SharePoint support (as a feature), with all credits given to the original one. However I need some kind of approval from you.

Huachao:
@s-KaiNet it seems publishing a new extension is a better way for now, and if you need any help, don't hesitate to leave comments here, I will keep this PR in case of you need any help or discussions.

flanakin:
@s-KaiNet I didn't create the VS Code app. That came from the VS Code team at Microsoft. I just reused the client id for auth. Some Microsoft apps are available in all directories, which is probably the case here.

Can't you leave the header decision up to the user? You could detect when it's on which one and react accordingly. This would be similar to how I use the endpoint to set the token's audience ...except a little more critical in your case :-P

s-KaiNet:
I'm closing this one in order not to bother you with unnecessary notifications.

PR #164, opened 2018-02-11
Request scripts

mkloubert:
scripts

I have implemented a new feature, which is able to run JavaScript code for a `.http` or `.rest` file (s.
 [request_scripts](https://github.com/mkloubert/vscode-restclient/tree/request_scripts) branch).

This makes it possible to realize batch operations, e.g.

**Example:**

You have a list of user IDs in a file like `users.txt`:

```
1
2
3
```

And a request like that:

```http
POST https://example.com/users/{{uid}}

{{body}}
```

You can run a JavaScript for each of that values:

```javascript
// load the list of user IDs
const USER_IDS = $fs.readFileSync($cwd + '/users.txt', 'utf8')
                    .split("\n")
                    .map(line => line.trim())
                    .filter(line => '' !== line);

$each(USER_IDS, (userId) => {
    // update progress text
    $progress("Loading data of user '" + userId + "' ...");

    // run a request for 'userId'
    return $request(
        // vars
        {
            // @uid
            'uid': userId,            
            // @body
            'body': $fs.readFileSync($cwd + '/user_to_create_' + userId + '.json')
        },

        // completed action
        (err, response) => {
            if (err) {
                $vscode.window.showErrorMessage('ERROR: ' + err);
            }
            else {
                $fs.writeFileSync($cwd + '/new_user_' + userId + '.json',
                                  response.body,
                                  'ascii');
            }
        }
    );
});
```

If you select `Run As REST Script` in the context menu from the editor of the script, the user is asked for the `.http` or `.rest` file, the script should be run for.

For more information see [current README](https://github.com/mkloubert/vscode-restclient/tree/request_scripts#scripts) file.


redcoreit:
This approach will be the official way of doing batch operations?
Scenario: 
- I have individual test files (.rest)
- I want to run all of them at once and check the minimalistic (only http code or something) result. Like a unit test... (I'm aware of the `'rest-client.previewOption': 'header'` option)

Huachao:
@mkloubert @redcoreit sorry for the late reply for this great PR, I am still considering about this feature. Previously, I just want to expose some hooks(_before request_ and _after response_) for user to add some script for special purpose. Like you can inject some custom variables, and save the response to some location.

src/controllers/requestController.ts
```diff
@@ -95,6 +123,138 @@ export class RequestController {
...
+                return;
+            }
+
+            const WORKSPACES = workspace.workspaceFolders;
+            const REQUEST_DOCUMENTS: StringDocument[] = [];
+
+            if (WORKSPACES) {
+                const FILES: string[] = [];
+                for (const WS of WORKSPACES) {
+                    (await glob('**/*.{http,rest}', {
```
Meir017: would it make sense to read the regex `.{http,rest}` from the `package.json` file https://github.com/Huachao/vscode-restclient/blob/master/package.json#L59?

ginomessmer:
+1 would love to see this feature.

PR #196, opened 2018-04-21
[Live Share] Limiting linting to local files

lostintangent:
This is a follow-up to #187, since I forgot to limit the variable linting to local files as well, and Live Share handles "remoting" diagnostics from the host to the guest.

// CC @Huachao


Huachao:
@lostintangent Merged, thanks.

Issue #205, opened 2018-05-03
Add support for default headers


Kronuz:
- VSCode Version: 1.22.2
- OS Version: macOS High Sierra 10.13.4
- REST Client Version: 0.18.3

Please add `rest-client.defaultHost`, `rest-client.defaultContentType`, `rest-client.defaultAccept`, etc.; so that, if configured, we don't have to specify `http://localhost:8080/` or `Host: localhost:8080`.

The idea is that for things like:

```http
PUT /documents/john
Host: localhost:8080
Content-Type: application/json
Accept: application/json; indent=4

{
    "firstname": "John",
    "lastName": "Smith"
}
```

One could simply configure the (workspace) settings:
```json
{
    "rest-client.defaultHost": "localhost:8080",
    "rest-client.defaultContentType": "application/json",
    "rest-client.defaultAccept": "application/json; indent=4"
}
```

and use:
```http
PUT /documents/john

{
    "firstname": "John",
    "lastName": "Smith"
}
```

The proposal is to convert settings of the form PascalCase to dash-separated (`defaultHeaderName` to `header-name`); so this would effectively extend `rest-client.defaultuseragent` (and rename it to `rest-client.defaultUserAgent`).

**Edit:**
I'd propose support for the following standard headers:

- Accept
- Accept-Charset
- Accept-Datetime
- Accept-Encoding
- Accept-Language
- Access-Control-Request-Headers
- Access-Control-Request-Method
- Authorization
- Cache-Control
- Content-Type
- Cookie
- Host
- Origin
- User-Agent

Huachao:
@Kronuz I have added some comment in #206, and what about 
```json
{
    "rest-client.defaultHeaders": {
         "X-Custom-Header": "foo bar",
         "Host": "localhost"
    }
}
```

PR #206, opened 2018-05-03
Default headers support

Kronuz:
This adds the feature in #205 by adding the following settings:

+ defaultAccept
+ defaultAcceptCharset
+ defaultAcceptDatetime
+ defaultAcceptEncoding
+ defaultAcceptLanguage
+ defaultAccessControlRequestHeaders
+ defaultAccessControlRequestMethod
+ defaultAuthorization
+ defaultCacheControl
+ defaultContentType
+ defaultCookie
+ defaultHost
+ defaultOrigin
+ defaultUserAgent

### DIFF
--- a/src/httpClient.ts
+++ b/src/httpClient.ts
@@ -81,9 +81,14 @@ export class HttpClient {
             options.headers = httpRequest.headers = {};
         }
 
-        // add default user agent if not specified
-        if (!hasHeader(options.headers, 'User-Agent')) {
-            options.headers['User-Agent'] = this._settings.defaultUserAgent;
+        // add default headers if not specified
+        for (let header in this._settings.defaultHeaders) {
+            if (!hasHeader(options.headers, header)) {
+                const value = this._settings.defaultHeaders[header];
+                if (value) {
+                    options.headers[header] = value;
+                }
+            }
         }
 
         let size = 0;
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ GET https://example.com/comments
 ```
 
 ### Request Headers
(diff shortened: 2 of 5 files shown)


Huachao:
@Kronuz Thanks for your PR, and I am considering that if we could directly provide a setting like `defaultHeaders`. And user just needs to place all their expected headers into this setting, even custom headers can be set. What's your opinion?

Kronuz:
That sounds also good... probably even better! Still, I’d get rid of `defaultuseragent` nonetheless, so there’s only one way to define that one.

One thing I stumbled upon while making the pull request is I had to create a new instance of the `RestClientSettings` in the `HttpRequestParser`, and i wasn’t very happy about it (as it adds new callbacks to and all)... it some other way you can think of to better access it from there?

Huachao:
I am also considering this question that in current code, every place needs to reference the settings, they have to create a new instance of the RestClientSettings class. My suggestion is that, it seems there's no need to hold many instances of setting class, can we just leverage the singleton pattern. That means add a new public static function which is responsible for creating a setting instance internally if not exists, otherwise return the created one. And the constructor of it can be even made private.

Kronuz:
@Huachao, I modified and added what you suggested to this pull request (defaultHeaders) 👍

src/httpClient.ts
```diff
@@ -83,8 +83,13 @@ export class HttpClient {
         }
 
         // add default user agent if not specified
```
Huachao: Please update the comment as well.

Huachao:
@Kronuz please also update your code related to settings, since I have used the singleton pattern for Settings class. Thanks for your contribution.

Huachao:
@Kronuz Merged, Thanks!

Kronuz:
Great! will you make a release any time soon? I can't wait to use the official extension again :)

Kronuz:
It’s ok, thank you! 👍

Huachao:
@Kronuz you could test the latest version 0.19.0 to experience this feature, thanks for your contribution.

aktxyz:
Can variables be used in the settings file for defaultHeaders ??

PR #233, opened 2018-08-15
Switch JSON formatting from 'jsonc-parser' to 'js-beautify'

ygraber:
The existing responseFormatUtility.ts makes use of 'jsonc-parser' to format incoming JSON. The speed doesn't vary much based on size of JSON, but it **does** vary depending on the number of "edits" that need to be made. When the current implementation tries to format JSON with many objects and sub objects, 30+ seconds can go by (and the waiting spinner stalls). 'pretty-data' uses JSON.stringify(JSON.parse)) which is quick, but will break some invalid JSON where a key is listed twice. I have found that 'js-beautify' is quick and acceptable.

I used this link to test, it is both a large JSON payload and one that is slow to format using 'jsonc-parser' (I gave up waiting), returned in seconds with 'js-beautify':  http://www.vizgr.org/historical-events/search.php?format=json&begin_date=-3000000&end_date=20151231&lang=en

### DIFF
--- a/package.json
+++ b/package.json
@@ -558,6 +558,7 @@
     "highlight.js": "^9.6.0",
     "httpsnippet": "^1.16.5",
     "iconv-lite": "^0.4.15",
+    "js-beautify": "^1.7.5",
     "jsonc-parser": "^1.0.3",
     "jsonpath": "^1.0.0",
     "mime-types": "^2.1.14",
--- a/src/utils/responseFormatUtility.ts
+++ b/src/utils/responseFormatUtility.ts
@@ -1,18 +1,16 @@
 'use strict';
 
-import { applyEdits, format as JSONFormat } from "jsonc-parser/lib/umd/main";
-import { EOL } from "os";
 import { window } from 'vscode';
 import { MimeUtility } from './mimeUtility';
 const pd = require('pretty-data').pd;
+const beautify = require('js-beautify').js_beautify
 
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -558,6 +558,7 @@
     "highlight.js": "^9.6.0",
     "httpsnippet": "^1.16.5",
     "iconv-lite": "^0.4.15",
+    "js-beautify": "^1.7.5",
     "jsonc-parser": "^1.0.3",
```
Huachao: Remove this package in dependencies

ygraber:
Done!

src/utils/responseFormatUtility.ts
```diff
@@ -1,18 +1,16 @@
...
+const beautify = require('js-beautify').js_beautify;
 
 export class ResponseFormatUtility {
     public static formatBody(body: string, contentType: string, suppressValidation: boolean): string {
         if (contentType) {
             if (MimeUtility.isJSON(contentType)) {
                 if (ResponseFormatUtility.IsJsonString(body)) {
-                    const edits = JSONFormat(body, undefined, { tabSize: 2, insertSpaces: true, eol: EOL });
-                    body = applyEdits(body, edits);
+                    body = beautify(body, { indent_size: 4 });
```
deanchen: Why did you increase the indent_size from 2 to 4 here?

Issue #300, opened 2018-12-05
Copy result body


asmith3006:
Along with "Save Response Body" in the top bar a "Copy Response Body" would be really helpful.

viktor-evdokimov:
So this is huge. copying result body was working for me and stopped working lately, really ruined my workflow =(

Huachao:
@viktor-evdokimov I haven't implemented this feature, what do you mean `was working for you`?

viktor-evdokimov:
Up to some point i was able to select text from result tab and copy it to clipboard, and that no longer works

viktor-evdokimov:
Created PR for this #317

Huachao:
@viktor-evdokimov what's your VSCode and my extension versions respectively? Since I can't repro this on my machine.

viktor-evdokimov:
I am on insider build of vscode and latest version of the plugin.

Huachao:
@viktor-evdokimov this is an known issue of Microsoft/vscode#65452, I will keep track of it. BTW, I add a comment in your PR, please help to resolve it when available. Thanks in advance.

Huachao:
@asmith3006 @Baoyx007 @pranayrauthu @michelemauro Thank for the help of @viktor-evdokimov , we have fixed this issue and will publish this feature in next release. And @viktor-evdokimov I will keep track of the issue you've mentioned.

PR #317, opened 2019-01-25
add copy-reponse-body [#300]

viktor-evdokimov:
Fixes issue #300 

Added button to copy response body since CMD+C does not work for me in the latest version


viktor-evdokimov:
Also I can't express how much time this extension have saved me! 🎉

Huachao:
@viktor-evdokimov Merged, thanks 😄

PR #384, opened 2019-07-08
added support for GraphQL

ferronrsmith:


src/utils/httpRequestParser.ts
```diff
@@ -100,7 +114,16 @@ export class HttpRequestParser implements IRequestParser {
...
-        if (this._restClientSettings.formParamEncodingStrategy !== FormParamEncodingStrategy.Never && body && typeof body === 'string' && MimeUtility.isFormUrlEncoded(contentTypeHeader)) {
+        if (isGraphQlRequest) {
+            variables = HttpRequestParser.parseRequestBody(variableLines, requestAbsoluteFilePath, contentTypeHeader);
+
+            let graphQlPayload = {
+                query: body,
+                variables: JSON.parse(variables.toString())
+            };
+            body = JSON.stringify(graphQlPayload, null, 2);
+            bodyLines = body.split(EOL);
```
Huachao: Use the `bodyLines` parsed in line 86, and in this case, no need to prettify the graphQlPayload in line 124?

ferronrsmith: `bodyLines` here is a bit different. since your actually converting `graphQlPayload`, which is different than what's on line 86

Huachao: I know your point. However, the `bodyLines` variable is only used as a parameter in `HttpRequest` constructor which stands for the `rawBody`. It is used to record the original body lines in request

ferronrsmith: no need to `prettify`, just came in handy when debugging GraphQL requests.

src/utils/httpRequestParser.ts
```diff
@@ -100,7 +114,15 @@ export class HttpRequestParser implements IRequestParser {
         // parse body
         const contentTypeHeader = getHeader(headers, 'content-type') || getHeader(this._restClientSettings.defaultHeaders, 'content-type');
         body = HttpRequestParser.parseRequestBody(bodyLines, requestAbsoluteFilePath, contentTypeHeader);
-        if (this._restClientSettings.formParamEncodingStrategy !== FormParamEncodingStrategy.Never && body && typeof body === 'string' && MimeUtility.isFormUrlEncoded(contentTypeHeader)) {
+        if (isGraphQlRequest) {
+            variables = HttpRequestParser.parseRequestBody(variableLines, requestAbsoluteFilePath, contentTypeHeader);
+
+            let graphQlPayload = {
+                query: body,
+                variables: JSON.parse(variables.toString())
```
Huachao: If no variables provided, do we still need to serialize this, or if we included `variables: null` in the payload, will the GraphQL server allow this?

ferronrsmith: yes, it still needs to be sent. I've never personally used a request without variables, but I'll add a default state

src/utils/httpRequestParser.ts
```diff
@@ -119,7 +119,7 @@ export class HttpRequestParser implements IRequestParser {
 
             let graphQlPayload = {
                 query: body,
-                variables: JSON.parse(variables.toString())
+                variables: variables ? JSON.parse(variables.toString()) : {}
```
Huachao: Or the variables should be omitted?

ferronrsmith: variables can't be omitted. Some servers have schema validations, that check for both `{query,variables}`

src/utils/httpRequestParser.ts
```diff
@@ -100,7 +114,15 @@ export class HttpRequestParser implements IRequestParser {
         // parse body
         const contentTypeHeader = getHeader(headers, 'content-type') || getHeader(this._restClientSettings.defaultHeaders, 'content-type');
         body = HttpRequestParser.parseRequestBody(bodyLines, requestAbsoluteFilePath, contentTypeHeader);
-        if (this._restClientSettings.formParamEncodingStrategy !== FormParamEncodingStrategy.Never && body && typeof body === 'string' && MimeUtility.isFormUrlEncoded(contentTypeHeader)) {
+        if (isGraphQlRequest) {
+            variables = HttpRequestParser.parseRequestBody(variableLines, requestAbsoluteFilePath, contentTypeHeader);
+
+            let graphQlPayload = {
+                query: body,
+                variables: variables ? JSON.parse(variables.toString()) : {}
```
Huachao: Or the variables should be omitted if variables not provided? I think the variables field is optional

ferronrsmith: Sorry, my comment got deleted. Some GraphQL servers do schema validation checks for `{query,variables}`. I'd rather send it, even if it's empty.

Huachao: Got

Huachao:
@ferronrsmith Merged, thanks

PR #395, opened 2019-07-24
Add $base64 variable

piotrkubisa:
I have added new `$base64` variable that accepts second parameter as a value that should be encoded using base64. It might be useful, when composing `application/json` requests with a nested stringified JSON inside one key of the payload.

Example

```console
GET http://localhost/
Content-Type: application/json

{
    "data": "{{$base64 {"hello": "World"} }}"
}
```

```console
GET / HTTP/1.1
Content-Length: 42
Content-Type: application/json
User-Agent: vscode-restclient

{
    "data": "eyJoZWxsbyI6ICJXb3JsZCJ9"
}
```


piotrkubisa:
I am closing, because I realized myself it does not satisfy my needs. Feel free to reopen it, if you will think it's interesting feature to have.

calloncampbell:
What about supporting URL encoding for query params?

PR #397, opened 2019-07-26
ci: reduce the size of the extension (13x improvement on install folder size)

badre429:
steps:
1. refactore package.json and move dependencies to devDependencies because after using webpack all  project dependencies  are bundled inside the final build 
2. refactore gif and png file on readme.md to use github raw files 

**Old install directory:** 
> size:32,3 MB (33 926 867 bytes) 
> size on disk:37,6 MB (39 493 632 bytes) 
> contains :3 788 Files, 423 Folders 

**New install directory:**
> size:2,66 MB (2 800 177 bytes)
> size on disk:2,70 MB (2 838 528 bytes)   
> contains :21 Files, 6 Folders

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -12,3 +12,9 @@ tslint.json
 vsc-extension-quickstart.md
 package-json.lock
 webpack.config.js
+images/*.gif
+images/curl-request.png
+images/http.png
+images/request-history.png
+images/navigate.png
+images/curl-request.png
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ content-type: application/json
     "time": "Wed, 21 Oct 2015 18:27:50 GMT"
 }
 ```
-Once you prepared a request, click the `Send Request` link above the request, or use shortcut `Ctrl+Alt+R`(`Cmd+Alt+R` for macOS), or right-click in the editor and then select `Send Request` in the menu, or press `F1` and then select/type `Rest Client: Send Request`, the response will be previewed in a separate __webview__ panel of Visual Studio Code. If you'd like to use the full power of searching, selecting or manipulating in Visual Studio Code, you can also preview response in __an untitled document__ by setting `rest-client.previewResponseInUntitledDocument` to `true`, by default the value is `false`. When a request is issued, ![cloud upload](images/loading.gif) will be displayed in the status bar, after receiving the response, the icon will be changed to the duration and response size.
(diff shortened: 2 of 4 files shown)


Huachao:
@badre429 thanks for your effort to reduce the extension size, however, I think moving the dependencies to devDependencies has no benefit to reducing the extension size after I exclude the `node_modules` in `.vscodeignore` file. In previous releases, the `node_modules` are included in the ultimate extension package. I suggest to only keep the image update part.

badre429:
@Huachao  
when you use webpack to build extension you have moving the dependencies to devDependencies for two reason 
1. webpack bundle all dependencies  in one file extension.js
2. vsce builder add all dependencies in package.json to output extension.vsxi file

Huachao:
@badre429 I am still not quite sure about your second point, I think that all the dependencies are already bundled into the extension.js, while, when creating a vsix file, it will take the `.vscodeingore` file into consideration, and I already excluded the `node_modules` in this file (https://github.com/Huachao/vscode-restclient/blob/36227f8ab8fa9ce99d4b5cd4e43ecf9856d99e4f/.vscodeignore#L3). And this change hasn't been published in any previous releases. In my local test, if I excluded the `dependency` part, the extension size is still around 3MB

badre429:
@badre429  
Despite the fact that you add 'node_modules' to .vscodeignore file you still end up with 'node_modules' inside the final directory unless you move  dependencies to devDependencies in this case you will no longer get 'node_modules' in the final output

Huachao:
> @badre429
> Despite the fact that you add 'node_modules' to .vscodeignore file you still end up with 'node_modules' inside the final directory unless you move dependencies to devDependencies in this case you will no longer get 'node_modules' in the final output
> 
> > you can test moving dependencies to devDependencies and compare the difference

Are you sure that the `node_moduels` directory is still in the final output? In my test, I can confirm that no `node_modules` directory will be generated:

1. Use `vsce package` command to generate the `.vsix` package
2. Go to the `Extensions` view of VSCode, and click the **...** button on the upper right corner, and select the `Install from VSIX` and then choose  generated `.vsix` file in the previous step.

After installation, check the folder structure under your extension install path, like `~/.vscode/extensions/`, you can find the *humao.rest-client -0.xxx* folder, and you can check the content there.

badre429:
@Huachao 
kept only image file

Huachao:
@badre429 Merged, thanks!

Issue #396, opened 2019-07-25
Request name as Tab title


lochstar:
Hi,

I sometimes have a number of request tabs open and they all have the generic title `Response(XXXms)`. Adding an option to use the `requestName` as the tab title, this would be helpful to identify which request is which.

I suggest something like the following:

### New Setting

`rest-client.requestNameAsTabTitle`: Show request name as the tab title. (Default is false)

### Example Request

```http
# @name get_todo
GET https://jsonplaceholder.typicode.com/todos/1
Content-Type: application/json
```

### Resulting Tab Title

![named response](https://i.imgur.com/Y0wT2ox.png)

This could potentionally be expanded to allow other request variables to be set as the title.

I'm happy to submit a PR to handle this feature, just thought I'd see if you were happy with this approach first.

Huachao:
@lochstar what's the behavior when a request has no name defined in this case?

lochstar:
Same as now, would just show `Response(XXXms)`.

Huachao:
@lochstar OK, and if you want to contribute this project, please feel free to submit a PR, thanks.

Huachao:
@lochstar this feature has been published in the latest version 0.22.1

lochstar:
Awesome! Cheers.

PR #400, opened 2019-08-01
Add setting to display Request Name as Tab Title

lochstar:
As outlined in #396, this works in conjunction with `showResponseInDifferentTab`.


README.md
```diff
@@ -585,6 +585,7 @@ exchange | Preview the whole HTTP exchange(request and response)
 * `rest-client.defaultHeaders`: If particular headers are omitted in request header, these will be added as headers for each request. (Default is `{ "User-Agent": "vscode-restclient", "Accept-Encoding": "gzip" }`)
 * `rest-client.timeoutinmilliseconds`: Timeout in milliseconds. 0 for infinity. (Default is __0__)
 * `rest-client.showResponseInDifferentTab`: Show response in different tab. (Default is __false__)
+* `rest-client.requestNameAsTabTitle`: Show request name as the tab title. (Default is __false__)
```
Huachao: How about renamed as `rest-client.requestNameAsResponseTabTitle`

package.json
```diff
@@ -351,6 +351,12 @@
           "scope": "resource",
           "description": "Show response in different tab"
         },
+        "rest-client.requestNameAsTabTitle": {
+          "type": "boolean",
+          "default": false,
+          "scope": "resource",
+          "description": "Show request name as the tab title."
```
Huachao: Add some description of the actual behavior when no request name is specified.

src/views/httpResponseWebview.ts
```diff
@@ -45,12 +45,13 @@ export class HttpResponseWebview extends BaseWebview {
         this.context.subscriptions.push(commands.registerCommand('rest-client.unfold-response', () => this.unfoldResponseBody()));
     }
 
-    public async render(response: HttpResponse, column: ViewColumn) {
+    public async render(response: HttpResponse, request: HttpRequest, column: ViewColumn) {
```
Huachao: There are two kinds of views of previewing responses, one is the `webview` here, the other is the `untitled document view`, does this setting applies in both views?

src/controllers/requestController.ts
```diff
@@ -149,7 +149,7 @@ export class RequestController {
                 if (this._restClientSettings.previewResponseInUntitledDocument) {
                     this._textDocumentView.render(response, previewColumn);
                 } else {
-                    this._webview.render(response, previewColumn);
+                    this._webview.render(response, httpRequest, previewColumn);
```
Huachao: No need to pass the `httpRequest` argument, since the underlying request already set in the response object. The only problem is that, the value of `request` field in response object doesn't set the `requestVariableCacheKey` value, so I suggest to set in the https://github.com/Huachao/vscode-restclient/blob/master/src/utils/httpClient.ts#L106

src/views/httpResponseWebview.ts
```diff
@@ -45,12 +45,13 @@ export class HttpResponseWebview extends BaseWebview {
...
         let panel: WebviewPanel;
         if (this.settings.showResponseInDifferentTab || this.panels.length === 0) {
+            const tabTitle = this.settings.requestNameAsTabTitle && request.requestVariableCacheKey && request.requestVariableCacheKey.key ? request.requestVariableCacheKey.key : 'Response';
```
Huachao: Simply
```js
this.settings.requestNameAsTabTitle && request.requestVariableCacheKey && request.requestVariableCacheKey.key || 'Response'
```

src/controllers/requestController.ts
lochstar: Thanks, will check it out.

src/views/httpResponseWebview.ts
lochstar: Doesn't apply for `untitled document view`, updated the README accordingly.

lochstar:
I've made the suggested changes and also fixed it so the tab title is set when _not_ displaying in a new tab. Let me know if there's anything else.

Huachao:
@lochstar Merged, thanks.

Issue #342, opened 2019-03-12
Allow environment variables to reference shared environment variables


ariehf:
Allow environment variables to reference shared environment variables.
This is useful when there are a large amount of environments.
(can choose to be implemented with or without using the "$shared " prefix)

For example multiple environments where half of the environments use one shared variable, and the other half uses a different shared variable.

    "rest-client.environmentVariables": {
        "$shared": {
            "tokenDev": "Bearer xyz...",
            "tokenProd": "Bearer abc..."
        },
        "webserverDev1": {
            "token": "{{$shared tokenDev}}",
            "host": "http://webserverDev1",
        },
        "webserverDev2": {
            "token": "{{$shared tokenDev}}",
            "host": "http://webserverDev2",
        },
        "webserverProd1": {
            "token": "{{$shared tokenProd}}",
            "host": "http://webserverProd1",
        },
        "webserverProd2": {
            "token": "{{$shared tokenProd}}",
            "host": "http://webserverProd2",
        }
    }

GET {{host}}/myendpoint
Authorization: {{token}}

uriel-sf:
👍

uriel-sf:
I have a related use case. I'd like to reference different shared variables for different environments. For example:
```
{
  "rest-client.environmentVariables": {
    "$shared": {
        "devApiUrl": "dev URL",
        "qaApiUrl": "qa URL",
    },
    "dev.user.multirate": {
      "apiUrl": "{{$shared devApiUrl}}",
    },
    "dev.user.perpetual": {
      "apiUrl": "{{$shared devApiUrl}}",
    },
    "qa.user.multirate": {
         "apiUrl": "{{$shared qaApiUrl}}"
     }
```

snackb:
Mind if I take a look at this?

Huachao:
@ariehf @uriel-sf @snackb with the great PR from @snackb this feature will be published in next release.

Huachao:
@ariehf @uriel-sf @snackb you can try this feature implemented by @snackb in the latest version 0.2.1

PR #409, opened 2019-08-21
Allow environment variables to reference shared variables

snackb:
Per issue #342, changes environment variable provider to check environment variables for values like {{$shared [xyz]}}, and if found attempts to get shared variable with name [xyz]. 

Added new error for trying to find a shared variable and not finding one. 

Added to README to reflect new functionality. 

This is my first contribution so I'm very open to corrections/advice!

### DIFF
--- a/src/utils/httpVariableProviders/environmentVariableProvider.ts
+++ b/src/utils/httpVariableProviders/environmentVariableProvider.ts
@@ -65,7 +65,7 @@ export class EnvironmentVariableProvider implements HttpVariableProvider {
                 continue;
             }
             const referenceKey = match[1].trim();
-            if (referenceKey in shared && typeof(shared[referenceKey]) === "string") {
+            if (typeof(shared[referenceKey]) === "string") {
                 current[key] = shared[referenceKey];
             }
         }
--- a/src/utils/httpVariableProviders/environmentVariableProvider.ts
+++ b/src/utils/httpVariableProviders/environmentVariableProvider.ts
@@ -54,10 +54,10 @@ export class EnvironmentVariableProvider implements HttpVariableProvider {
         return Object.assign({}, sharedEnvironmentVariables, currentEnvironmentVariables);
     }
 
-    private mapEnvironmentVariables(current: any, shared: any){
+    private mapEnvironmentVariables(current: any, shared: any) {
         for (let key in current) {
             const value = current[key];
-            if (!(typeof(value) === "string")) {
+            if (typeof(value) !== "string") {
                 continue;
             }
             const variableRegex = /\{{2}\$shared (.+?)\}{2}/;
@@ -67,7 +67,7 @@ export class EnvironmentVariableProvider implements HttpVariableProvider {
             }
             const referenceKey = match[1].trim();
             const referenceValue = shared[referenceKey];
-            if(!referenceValue) {
+            if (!referenceValue && referenceValue !== "") {
                 continue;
             }
             current[key] = referenceValue;
--- a/src/utils/httpVariableProviders/environmentVariableProvider.ts
+++ b/src/utils/httpVariableProviders/environmentVariableProvider.ts
@@ -13,6 +13,7 @@ export class EnvironmentVariableProvider implements HttpVariableProvider {
     private readonly _settings: RestClientSettings = RestClientSettings.Instance;
 
     public static get Instance(): EnvironmentVariableProvider {
+
         if (!EnvironmentVariableProvider._instance) {
             EnvironmentVariableProvider._instance = new EnvironmentVariableProvider();
         }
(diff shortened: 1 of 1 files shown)


src/utils/httpVariableProviders/environmentVariableProvider.ts
```diff
@@ -44,11 +53,25 @@ export class EnvironmentVariableProvider implements HttpVariableProvider {
         return Object.keys(variables).map(key => ({ name: key, value: variables[key]}));
```
Huachao: `GetAll` method should also return the actual value for referenced variable as `Get`

src/utils/httpVariableProviders/environmentVariableProvider.ts
```diff
@@ -44,11 +53,25 @@ export class EnvironmentVariableProvider implements HttpVariableProvider {
...
 
+    private getShared(name: string): HttpVariable {
+        const variables = this.getSharedVariables();
+        if (!(name in variables)) {
+            return { name, error: ResolveErrorMessage.EnvironmentReferenceVariableNotExist };
+        }
+        return { name, value: variables[name] };
+    }
+
     private async getAvailableVariables(): Promise<{ [key: string]: string }> {
```
Huachao: I suggest that `getAvailableVaraibles` returns the resolved variables. And for non-exist variables, no error is emitted, just leave it as is. And in this case, `getShared` and `getSharedVariables` are no longer needed.

src/utils/httpVariableProviders/environmentVariableProvider.ts
```diff
@@ -49,6 +50,27 @@ export class EnvironmentVariableProvider implements HttpVariableProvider {
...
         const currentEnvironmentVariables = variables[environmentName];
         const sharedEnvironmentVariables = variables[EnvironmentController.sharedEnvironmentName];
+        this.mapEnvironmentVariables(currentEnvironmentVariables, sharedEnvironmentVariables);
         return Object.assign({}, sharedEnvironmentVariables, currentEnvironmentVariables);
     }
+
+    private mapEnvironmentVariables(current: any, shared: any) {
+        for (let key in current) {
+            const value = current[key];
+            if (!(typeof(value) === "string")) {
```
Huachao: typeof(value) !== 'string'

src/utils/httpVariableProviders/environmentVariableProvider.ts
```diff
@@ -49,6 +50,27 @@ export class EnvironmentVariableProvider implements HttpVariableProvider {
...
+                continue;
+            }
+            const variableRegex = /\{{2}\$shared (.+?)\}{2}/;
+            const match = variableRegex.exec(value);
+            if (!match) {
+                continue;
+            }
+            const referenceKey = match[1].trim();
+            const referenceValue = shared[referenceKey];
```
Huachao: If the value defined in the `$shared` environment is an empty string. in this case, it will be ignored

snackb:
Apologies, those were meant to be on a branch, I was going to have someone else look over them first this time!

In any case, thank you very much for your patience.

src/utils/httpVariableProviders/environmentVariableProvider.ts
```diff
@@ -49,6 +50,27 @@ export class EnvironmentVariableProvider implements HttpVariableProvider {
         const variables = this._settings.environmentVariables;
         const currentEnvironmentVariables = variables[environmentName];
         const sharedEnvironmentVariables = variables[EnvironmentController.sharedEnvironmentName];
+        this.mapEnvironmentVariables(currentEnvironmentVariables, sharedEnvironmentVariables);
         return Object.assign({}, sharedEnvironmentVariables, currentEnvironmentVariables);
     }
+
+    private mapEnvironmentVariables(current: any, shared: any) {
+        for (let key in current) {
```
Huachao: minor: Use for...of together with `Object.entries` since you need to use both key and value pair like following:
```js
for (const [key, value] of Object.entries(current)) {
    // Logic
}
```

src/utils/httpVariableProviders/environmentVariableProvider.ts
```diff
@@ -49,6 +50,27 @@ export class EnvironmentVariableProvider implements HttpVariableProvider {
...
+            if (typeof(value) !== "string") {
+                continue;
+            }
+            const variableRegex = /\{{2}\$shared (.+?)\}{2}/;
+            const match = variableRegex.exec(value);
+            if (!match) {
+                continue;
+            }
+            const referenceKey = match[1].trim();
+            const referenceValue = shared[referenceKey];
```
Huachao: How about directly using `in`  operator to test if the given key in object like following:
```js
const referenceKey = match[1].trim();
if (referenceKey in shared) {
    current[key] = shared[referenceKey];
}
```

snackb: Alright I've made these changes. However I've just noticed this doesn't update the breadcrumbs yet:
image
I'll work on this now.

Huachao:
@snackb Merged, thanks 😄

snackb:
And thank you!

PR #434, opened 2019-10-21
Add JMESPath support

mburleigh:
Add support for piping the response body through a JMESPath query to pull a subset of data from the response.

For example, to see the title & created date for all issues in this repo it would look like

https://api.github.com/repos/huachao/vscode-restclient/issues 

| [].{Title:title,Created:created_at}


Huachao:
@mburleigh I think the JMES path support doesn't follow the principle of my extension. The introduce of `JSONPath` and `XPath` are used for extracting the expected response part for subsequent requests, and leave the raw response as-is.  For example, if a login request returns a response with a token and other requests need it, I will allow the `request varaible` to pick the exected token with JSON path. So personally, I just think JMES path is a counterpart with JSONPath.

mburleigh:
Hi @Huachao ; thanks for this great extension and for looking at my PR. I use the request variable support sometimes (especially for auth tokens) but I think JMESPath is complementary to JSONPath. I see some overlap (slicing, filtering, etc) but I don't see a JSONPath analog of projection or the functions built in to JMESPath. Another difference is the intent; I think JMESPath works best to pull out the interesting part of the response and display it while the JSONPath is better when used in the context of a request variable.

Huachao:
@mburleigh I agree with you that JMESPath provides extra ability to extract the JSON response. And I think that we can add the JMESPath support in the request variable. However, I don't want to directly embedded the JMESPath support to update the received response. Previously, I have an idea that adding an input box in the response preview panel, and as user types the JSONPath/JMESPath, the original rendered response will be modified

mburleigh:
We are a lot of time zones apart :)

There are already extensions for Code that add an input box that you can type a JMESPath query into that will be executed against the focused editor and show results in the Output window (https://marketplace.visualstudio.com/search?term=jmespath&target=VSCode&category=All%20categories&sortBy=Relevance).

I like the idea of being able to save the queries in the file since they can get long and complicated. What would you think about opening a second result editor with the JMESPath query results? (that would preserve the original response) Or maybe a preference to preserve (or not) the response when  a JMESPath query is detected?

Huachao:
@mburleigh thanks for your response, actually my idea is a bit different from what you think, I mean just add JSONPath input at the bottom of the preview panel, and if you type anything, the previewed response will be reflected on-the-fly, if you clear the input, the previewed response will back to original one.

However, I think `JSONPath`, `XPath` and `JMESPath` are equivalent. Current JSONPath support is used for constructing the request variables, instead of the above idea.

mburleigh:
Ok, maybe I'm not aware of the full functionality of the JSONPath support. To be clear I am not suggesting JMESPath as a replacement for JSONPath or the way the request/response variables are being handled.

The use case I have in mind is this:
Given a response that contains a JSON array of hundreds of items I want to see only a subset of the items (imagine the list is orders and I want the orders for a single customer). Assuming I have a request variable and can craft the appropriate JSONPath expression to filter the array items how can I then view the filtered list? I know if I hover over the JSONPath expression I will see the value in a tooltip (very useful for auth tokens); is the any way to put the JSONPath result in a tab so it can be easily inspected? 

That would do most of what I hoped to do with this PR.

Huachao:
@mburleigh thanks for your response. I know your point, and what I thought is the #7 feature request, and JSONPath or JMESPath are just ways to achieve it.

mburleigh:
Yes, that is pretty much the feature I'm trying to implement here. Are you set on a panel that is part of the response view or would you be ok with something opening side by side with the response? I think you will find the idea of typing in the query manually every time is not sustainable; the queries (JMESPath at least) can get very involved when there is nesting, filtering, & projections involved. When that is the case it is very nice to have them saved somewhere.

What would you think about adding something that would allow a JMESPath query to be executed against an existing request variable? Maybe add an "Execute Query" CodeLens item that would take a request variable and a query and output the results in a new tab.

Huachao:
@mburleigh sorry for the late reply. Can we achieve what you described by `Save Response Body` in the response preview panel, and execute the JMES query with the JMES related extensions in the marketplace?

mburleigh:
@Huachao I hope you had a great weekend!

I have tried setting the Preview Option setting to 'body' and then using the existing JMESPath extensions to execute a query against the results. This works fine but has a few drawbacks:
* the query result is shown in the output window so there is no option to compare queries, collapse regions, limited copy/paste, etc (you lose all the power of a real editor tab)
* neither current extension has a provision to save a JMESPath query (and even if they could a query tends to be tightly coupled to the result so a list of saved queries would be less useful than having the queries saved alongside the requests)

Here is an example of a fairly complex query; you can see why manual entry would be painful:
```
[].{Id:pullRequestId,By:createdBy.displayName,Repo:repository.name,Source:sourceRefName,Target:targetRefName,Internal:reviewers[?contains(displayName, 'Internal')].vote | [0],External:reviewers[?contains(displayName, 'External')].vote | [0],Title:title,URL:join('', ['https://dev.azure.com/pandora-jewelry/PANDORA%20B2B%20Cloud/_git/', repository.name, '/pullrequest/', to_string(pullRequestId), '?_a=overview'])}
```

That query takes a JSON that has all the PRs for an Azure DevQps Team Project and shows which ones are still pending review (in addition to some other information). Since this is dependent on the shape of the JSON I think it would be preferable to save it in the same file as the request that returns the JSON.

mburleigh:
Hi @Huachao; looks like none of the options I presented for adding this functionality are going to be acceptable for you. As an alternative I've been thinking about an extension that extends the REST Client and adds a second CodeLens item ("Execute Query") next to the "Send Request" item contributed by REST Client. I've been trying to come up with a POC that just references your code but that gets really ugly when it comes to making an actual request.

Would you be open to some refactoring of the REST Client code to expose an API that other extensions could consume (https://code.visualstudio.com/api/references/vscode-api#extensions)? I think the changes would be primarily in the RequestController class. If you are ok with this I can give the refactoring a try.

PR #440, opened 2019-11-06
Replaced images with vscode-icons

lochstar:
I noticed the command images looked a little out-of-place since VSCode updated their icons.

I've replaced the images with the official [vscode-icons](https://github.com/microsoft/vscode-icons) and mapped them as follows:

Command | Icon Name
------------ | -------------
save | save
save-body | save-all
copy | files

Current:
current

New:
new

Happy to change any if you'd prefer an alternative.


Huachao:
@lochstar thanks, merged!

Issue #453, opened 2019-12-04
Support turn on/off retry request


chuongtrh:
I found a document from [got](https://github.com/sindresorhus/got#retry) with the **retry** option
Can you support this feature in the future?

chuongtrh:
hi @Huachao 
I have created a pull request [458](https://github.com/Huachao/vscode-restclient/pull/458)
Can you check it? 
Thanks!

PR #458, opened 2019-12-16
~ Support retry request

chuongtrh:
Base on a feature request [#453](https://github.com/Huachao/vscode-restclient/issues/453)

### DIFF
--- a/README.md
+++ b/README.md
@@ -624,6 +624,7 @@ exchange | Preview the whole HTTP exchange(request and response)
 * `rest-client.logLevel`: The verbosity of logging in the REST output panel. (Default is __error__)
 * `rest-client.enableSendRequestCodeLens`: Enable/disable sending request CodeLens in request file. (Default is __true__)
 * `rest-client.enableCustomVariableReferencesCodeLens`: Enable/disable custom variable references CodeLens in request file. (Default is __true__)
+* `rest-client.retry`: Number of retry request. (Default is __0__)
 
 Rest Client extension respects the proxy settings made for Visual Studio Code (`http.proxy` and `http.proxyStrictSSL`). Only HTTP and HTTPS proxies are supported.
 
--- a/package.json
+++ b/package.json
@@ -603,6 +603,12 @@
           "default": true,
           "scope": "resource",
           "description": "Enable/disable custom variable references CodeLens in request file"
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -624,6 +624,7 @@ exchange | Preview the whole HTTP exchange(request and response)
 * `rest-client.logLevel`: The verbosity of logging in the REST output panel. (Default is __error__)
 * `rest-client.enableSendRequestCodeLens`: Enable/disable sending request CodeLens in request file. (Default is __true__)
 * `rest-client.enableCustomVariableReferencesCodeLens`: Enable/disable custom variable references CodeLens in request file. (Default is __true__)
+* `rest-client.retry`: Number of retry request. (Default is __0__)
```
Huachao: It seems not clear for me that when will the retry operation occur if we set the value to larger than 0? I think the behavior should be documented

chuongtrh: Hi @Huachao

> The number limit of retry requests. The default is turned off with 0. 

Is it clear? Any advice for the document better. Thank

Huachao: @chuongtrh I mean as an extension user,  how can I know in which circumstance the retry operation will happen.

chuongtrh: @Huachao Base on the document from [got](https://github.com/sindresorhus/got#retry). When response with status **500**, the extension will be retry 2 times because the limit default is 2. I think it is good if extension has option setting retry limit.

Huachao: @chuongtrh are you sure **got**  package behaves as you described? https://github.com/sindresorhus/got/blob/master/readme.md#retry

chuongtrh: @Huachao yes. I make a simple request test like the image above

image
image

Huachao: @chuongtrh from the doc link I have posted, **500** is just one possible reason for retry happens

chuongtrh: @Huachao Correct. 
I found this situation when I try to write to the log . It will be writing 3 times when making a request via rest-client. It is hard when debugging. Play with postman only 1 time

This option is not must have but have is good :) Thank for the quick reply.

Huachao:
@chuongtrh as we still have some uncertainty about retry reason and debug difficulty, I'd like to leave this PR open and wait for the community feedback

Issue #444, opened 2019-11-13
Send all requests in a file


bm-stschneider:
If I have multiple requests in a file and they have request/response variables. It would be nice if I could run all the request in sequence so that all get executed based on the same initial request.

Huachao:
@bm-stschneider do you mean just run all the requests in the correct order and display all the responses

bm-stschneider:
pretty much yes but also with possible request variables

AZ2Tonez:
Gave a thumbs up, but also want to comment that I would love this feature as well. Would also like to be able to use it as part of CI script as mentioned in https://github.com/Huachao/vscode-restclient/pull/472.

saurus:
Thanks @AZ2Tonez, 

BTW, I want to clarify that my pull request was mostly a proof of concept, and not intended to be merged as is.

But I hoped that my pull request can be a starting point to talk about this feature. If maybe @Huachao  can tell us what he thinks about this topic, I can help coding things in an acceptable way.

swhillance-mphn:
Would love this feature as well. I only discovered this extension today and I've already got a need for such a feature. Using request variables too

0x49D1:
This really would be nice: run selected or run all in the file. For example now I'm testing the REST API using the client and just need to run all and recheck the results.

boboz2:
+1 for this feature

Kiyozz:
+1

visschersm:
I really like the REST client to test my Requests, but adding this feature will make me close my postman application completely.

zhmq0124v:
+1

ricmrodrigues:
was this implemented?

Huachao:
> was this implemented?

@ricmrodrigues No, it wasn't.

chadbr:
@Huachao any interest in a PR?

jplonghi:
+1

api1st:
https://github.com/api1st/httprun is a Windows/Linux command-line tool for executing the .http files that vscode-restclient uses.

andreifloroiu:
why wouldn't this be a feature?

Trolldemorted:
>https://github.com/api1st/httprun is a Windows/Linux command-line tool for executing the .http files that vscode-restclient uses.

Unfortunately httprun is abandonware and still runs on dotnet5.

Issue #291, opened 2018-11-08
Automatic runs


fleed:
Is there a way to automatically run the saved requests and collect the results?

I'd like to integrate the requests in build/release pipelines for integration tests

PS: I know that it is a little bit out of scope, but it would be the definitive companion for the client. Additionally, the runner could be embedded in VS code as additional tab and be used to quickly run and monitor a set of requests

ishepherd:
+1, this could let us ditch Postman/Newman
How would one run such a thing from command line? npm?

ishepherd:
This is a dupe of https://github.com/Huachao/vscode-restclient/issues/267

lonix1:
> +1, this could let us ditch Postman/Newman

How cool would that be - to have one less app open during development!

Issue #267, opened 2018-10-10
How to write tests on top of the requests?


dhmuralikrishna:
How can we write acceptance tests using rest clients. Lets say if i have 50 request how can i test all those 50 calls are success. with some test cases.

dhmuralikrishna:
How to execute all http calls from cli.

Huachao:
@dhmuralikrishna nice suggestion, I think this feature is quite useful, and I will consider it carefully. Thanks

dhmuralikrishna:
Compared to postman rest client is very flexible.  Don't need to jump in multiple tabs.You did a great job. With postman collection and newman we can run the test through cli. it would be nice if we do the same with rest client.

Huachao:
@dhmuralikrishna nice suggestion, I'd like to keep this request open and wait to see if others vote this up.

jgroom33:
Here's a good project to model:
https://github.com/eykrehbein/strest

I'd like to see the benefits of this project and strest joined in a single project or have this project leverage an api of strest.

faisalraja:
This would be great. I'm thinking something like another divider for assert results:

GET ....

>>>

assert(ContentType == "application/json")
const data = JSON.parse(body)
assert(data.result.length == 5)

###

It would just have an additional data when running it on its own with pass fail. But running it for the whole file would give pass or fail for each. 

Plus a compatible cli that can run a set of files.

This has been my go to rest client. Thank you

jgroom33:
I've switched to using strest and wrote this plugin: https://github.com/jgroom33/vscode-strest-client

ishepherd:
> assert {{response.headers.ContentType == "application/json"}}

+1 looks more flexible/natural than strest's approach where you code your validations as an AST.

mekwall:
Where is the parser located for the http language? If that could be moved out to a separate package it would be easy to contribute with tooling for testing.

faisalraja:
I created a very specific needs that has very similar format with vscode-restclient since I like it's approach. It's still incomplete since we're only using this in our project and we add features as needed. [testhttp](https://github.com/faisalraja/testhttp) will also run dependencies, for example you run a test it will run all required calls to populate the variable plus importing a different file to use it's variables too.

raedwa01:
I would also like to see a way to implement these tests in a CI system like Jenkins.

calebjenkins:
@Huachao  - It would be worth looking at how httpYac does assertions. 
[...]

Issue #238, opened 2018-08-17
[Feature Request] Support Response Scripting


loverde:
I think this is a consolidation of #86, #181, and #209, but it requests a variant that would be compatible with JetBrain's REST Client http file definitions.

https://www.jetbrains.com/help/idea/http-client-in-product-code-editor.html

E.g.

```
GET /someplace_to_get_a_token

> {%
    var value = response.body.json.access_token;  
    // or more complex javascript logic to parse the response
    client.global.set("access_token", value);
%}
```

and support for the response scripts being external files:
```
> scripts/extract_access_token.js
```

The JetBrains API also supports client.test and client.assert for running tests/assertions on the response.

Regarding scoping of variables, it seems that JetBrains puts these in a transient scope that overrides the global values. But I'm not even sure how they handle viewing and resetting those values (you might have to restart the IDE).

So maybe it would be simpler to store them in the workspace rest-client.environmentVariables location for the currently selected environment (assuming it's possible to change those values from the VS Code API). That also has the benefit of preserving them and having a place to view and change them easily.

Or you could have a different API from JetBrains (see previous edit of this comment), but it would be nice to try to remain compatible where possible. It's kind of ugly, but maybe have a rest-client.globalSetterScope setting that determines where the client.global values get saved to.

If #21 is implemented to allow running multiple requests in a file at once, this feature would make that feature significantly more useful as well.

loverde:
Looks like VSCode does support changing the config via the API:
https://github.com/Microsoft/vscode/blob/master/src/vs/vscode.d.ts#L3946

I guess the other option for storing as "transient" is to store it in ExtensionContext.workspaceState:
https://github.com/Microsoft/vscode/blob/master/src/vs/vscode.d.ts#L4510

Huachao:
@loverde nice suggestion, I considered the script support for a long time. I will update in this thread later.

werwolfby:
I love JetBrains REST client! Will be happy to have support here.

Do you need any help with the implementation?

EvgenyKhaliper:
Yeah i would love to be able to extract values from json and then use as variables in upcoming call. This is the only thing that missing for me in this great extension.

Huachao:
@EvgenyKhaliper your requirement is already well supported via request variables, or even file variables reference request variables. For more details, you can refer the README for more details

EvgenyKhaliper:
@Huachao sorry, i've forgot to mention 'json' is an output of other REST call.

[...]

PR #472, opened 2020-01-11
Vscode independence

saurus:
Hi, 

I'm using `vscode-restclient` to try some complex api. I find this tool easy to use, while allowing to connect multiple calls.

As a few other people out there, I would like to use it for api validation, executing complex api orchestrations, and batch running for CI.

There are some of the Issues asking for features related to this PR:

* Send all requests in a file #444 
* Automatic runs #291 
* How to write tests on top of the requests? #267 
* [Feature Request] Support Response Scripting #238 
* Execute the same request N time #211 
* could you please add some before/after script ? #209 
* Run all requests one after another does not work #199 

With this pull request, I'm trying to remove dependency from Visual Studio Code internals from the modules that can be reused to create a command line tool.

Before explaining the changes, bear in mind that this is the first time I write code in Typescript, so forgive me if I did some silly mistakes. 

To remove `vscode` dependencies, I have mainly used three strategies:
- for classes with few dependencies, I have splitted in two classes: a base class, without the features depending on `vscode`, and a derived class adding those features. for example, `RestClientSettings` was divided in the base `RestClientSettings` (keeeping the name to minimise renaming changes), and `RestClientSettingsVS`. A very important class modified this way is `RequestController`, almost all the code went to `RestClient` class, and this class have only the Selection and Range handling. To aid this separation, also for `Selector` class some code went to `SelectorVS`.  
-  the class `TextDocument` was used in some places. to isolate the dependency, a new interface named `DocumentWrapper` (yes, I know, not a great name...), was created, and implemented in `DocumentWrapperVS` using TextDocument.
- To create the right instance of a class, some factory was needed. To avoid changing too much, I took advantage of `RestClientSettings` class, already widely used. On this class I have put some methods to hide dependencies on `window`, and some helpers to get information from current document.

I couldn't find a way to remove `vscode` dependencies from `SystemVariableProvider`, so by default the code uses a `DummyVariableProvider`, and the real class is set at startup. Also the EventEmitter class used in somewhere was a bit of problem. I don't think we need those events on a command line interface, but not being able to remove the dependency, I've implemented an interface to hide the dependency.

To check that all this work enables us to create a command line program, I created a bare-bone program. See [restclient-cli](https://github.com/saurus/restclient-cli).

Ok, let me know what you think about all this.

Thanks,

Dino


eaglemoor:
Any update ?

Huachao:
@eaglemoor no more update, as I said in previous threads, I'd like to keep the extension in VS Code scope only for now

saurus:
As there is not interest on this topic, I'm closing the pull request.

PR #496, opened 2020-01-24
Implement shorthand syntax for localhost 

### LINKED ISSUE #481: Support port-only notation for localhost
I. e. `:4000/foo` for `http://localhost:4000/foo` and `:/bar` for `http://localhost/bar` – [HTTPie does this](https://httpie.org/doc#url-shortcuts-for-localhost) and some (me :) find it very convenient.

Thank you for this awesome project BTW!

notpushkin:
(closes #481)

### DIFF
--- a/src/utils/httpRequestParser.ts
+++ b/src/utils/httpRequestParser.ts
@@ -103,6 +103,17 @@ export class HttpRequestParser implements IRequestParser {
             requestLine.url = `${scheme}://${host}${requestLine.url}`;
         }
 
+        // shorthand syntax for localhost
+        if (requestLine.url[0] === ':') {
+            if (requestLine.url[1] === '/') {
+                requestLine.url = `http://localhost:80${requestLine.url.slice(1)}`;
+            } else {
+                const [port] = requestLine.url.slice(1).split("/", 1);
+                const scheme = port === '443' || port === '8443' ? 'https' : 'http';
+                requestLine.url = `${scheme}://localhost${requestLine.url}`;
+            }
+        }
+
         // parse body
         const contentTypeHeader = getContentType(headers) || getContentType(this._restClientSettings.defaultHeaders);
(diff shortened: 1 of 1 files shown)


src/utils/httpRequestParser.ts
```diff
@@ -103,6 +103,17 @@ export class HttpRequestParser implements IRequestParser {
             requestLine.url = `${scheme}://${host}${requestLine.url}`;
         }
 
+        // shorthand syntax for localhost
```
Huachao: Can we move this logic into the `parseRequestLine` method

src/utils/httpRequestParser.ts
```diff
@@ -103,6 +103,17 @@ export class HttpRequestParser implements IRequestParser {
             requestLine.url = `${scheme}://${host}${requestLine.url}`;
         }
 
+        // shorthand syntax for localhost
+        if (requestLine.url[0] === ':') {
+            if (requestLine.url[1] === '/') {
+                requestLine.url = `http://localhost:80${requestLine.url.slice(1)}`;
+            } else {
```
Huachao: How about the case for `:` only?

Huachao:
@notpushkin, as documented in HTTPie, the `:port` style, is a curl-like shorthand, however when I test with `cURL` locally with this style, it's not supported. The original pull is here https://github.com/jakubroztocil/httpie/pull/171

src/utils/httpRequestParser.ts
notpushkin: This works too! See the check in line 108: “if `:` is directly followed by `/` (i. e. the path part), treat it as `localhost:80`”.

Huachao: No, Imean the value of `requestLine.url` is `:`, in this case, the final value of it is `http://localhost:`

notpushkin: Ah, makes sense! I'll add the changes later today. :sparkles:

Huachao:
@notpushkin any updates?

PR #497, opened 2020-01-24
Allow aad resourceuri

pschaeflein:
When using the $aadToken variable, the audience parameter (aud) is optional. If provided, it must be a Guid.

When calling an endpoint that is [protected via AAD ](https://docs.microsoft.com/en-us/azure/active-directory/develop/scenario-protected-web-api-overview), the endpoint's registration has a [Resource URI](https://docs.microsoft.com/en-us/azure/active-directory/develop/scenario-protected-web-api-app-registration#expose-an-api). The documentation recommends that the URI uses the pattern `api://{clientId}`.  

Using this recommended pattern in the extension does not work. Since the `aud` parameter is not a guid, the request URL is used as the resource URI. 

This change allows the `aud` parameter to be a URI.


src/utils/httpVariableProviders/systemVariableProvider.ts
```diff
@@ -32,7 +32,7 @@ export class SystemVariableProvider implements HttpVariableProvider {
 
     private readonly requestUrlRegex: RegExp = /^(?:[^\s]+\s+)([^:]*:\/\/\/?[^/\s]*\/?)/;
 
-    private readonly aadRegex: RegExp = new RegExp(`\\s*\\${Constants.AzureActiveDirectoryVariableName}(\\s+(${Constants.AzureActiveDirectoryForceNewOption}))?(\\s+(ppe|public|cn|de|us))?(\\s+([^\\.]+\\.[^\\}\\s]+|[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}))?(\\s+aud:([^\\.]+\\.[^\\}\\s]+|[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}))?\\s*`);
+    private readonly aadRegex: RegExp = new RegExp(`\\s*\\${Constants.AzureActiveDirectoryVariableName}(\\s+(${Constants.AzureActiveDirectoryForceNewOption}))?(\\s+(ppe|public|cn|de|us))?(\\s+([^\\.]+\\.[^\\}\\s]+|[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}))?(\\s+aud:([^\\.]+\\.[^\\}\\s]+|[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}|([^:]*:\/\/\/?[^/\s]*\/?)))?\\s*`);
```
Huachao: @pschaeflein need to escape all the backslash in the string

PR #532, opened 2020-02-28
feat: Eval system variable

jnorkus:
A quick prototype of an Eval system variable that allows executing any process to calculate a value. This is something that would work in my usecase, maybe others would find it useful too.


jnorkus:
@Huachao if this is something you're interested in adding to the extension, I will make some improvements and add documentation.

darewreck54:
I need something like this.  Was there any response in adding this back to the plugin vs. having to check out a forked version?  @Huachao @jnorkus

jnorkus:
Unfortunately I have not heard anything back from the author. I'm using my own forked build.

kibiz0r:
@jnorkus  Big drawback with this: As you're editing, VSCode continually re-evaluates these vars, so you have to be really careful about what you put in a `{{ $eval ... }}` block even as you're typing each keystroke, or you might delete a bunch of files or forkbomb your system, etc.

For instance, don't start typing `{{ $eval rm -rf ~/.cache/whatever/foo }}` unless you wanna blow up your whole home dir.

Marvin-Brouwer:
@jnorkus could you perhaps rewrite the $eval to only accept local JS/TS files (local as in relative path, that doesn't leave the workspace)
This would make it somewhat secure, maybe keep track of what is allowed to execute like pnpm does when you add new packages.

Having some kind of eval support would fix my requirement for https://github.com/Huachao/vscode-restclient/issues/706
And possibly for: https://github.com/Huachao/vscode-restclient/issues/870

PR #541, opened 2020-03-09
Include "from .http and .rest file" in the summary

aeberhart:
The README is quite long and the key information on how to get started (.http and .rest file extensions support) is quite far down. I overlooked it at first.


Huachao:
@aeberhart I think it's a bit too detailed to include this supported file formats in the extension one-sentence description. But still thanks for your contribution, really appreciated,

Huachao:
@aeberhart I think it's a bit too detailed to include this supported file formats in the extension one-sentence description. But still thanks for your contribution, really appreciated,

PR #590, opened 2020-05-19
Add support for the XDG directory specification

tristan957:
I am a Linux user.

Linux users hate when people put dotfiles in the HOME folder.

I noticed your awesome extension <3 puts its folder in the HOME folder.

I implemented backwards-compatible support for the XDG directory specification[1].

In the case `~/.rest-client` exists, we continue to use that folder. Otherwise, we use both `XDG_CACHE_HOME/rest-client` and `XDG_CONFIG_HOME/rest-client`.

So in my case as a previous user, I would just delete the current folder, and have the extension use the paths because I have those variables defined, as do most Linux users.

1: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

### DIFF
--- a/src/utils/userDataManager.ts
+++ b/src/utils/userDataManager.ts
@@ -13,7 +13,7 @@ function getCachePath(): string {
     }
 
     if (process.env.XDG_CACHE_HOME !== undefined) {
-        return path.join(rootPath, process.env.XDG_CACHE_HOME);
+        return path.join(process.env.XDG_CACHE_HOME, restClientDir);
     }
 
     return rootPath;
@@ -25,7 +25,7 @@ function getConfigPath(): string {
     }
 
     if (process.env.XDG_CONFIG_HOME !== undefined) {
-        return path.join(rootPath, process.env.XDG_CONFIG_HOME);
+        return path.join(process.env.XDG_CONFIG_HOME, restClientDir);
     }
 
     return rootPath;
--- a/src/utils/userDataManager.ts
+++ b/src/utils/userDataManager.ts
@@ -4,30 +4,58 @@ import * as path from 'path';
 import { HistoricalHttpRequest } from '../models/httpRequest';
 import { JsonFileUtility } from './jsonFileUtility';
 
+const restClientDir = 'rest-client';
+const rootPath = path.join(os.homedir(), `.${restClientDir}`);
+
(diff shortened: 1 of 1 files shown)


tristan957:
I cannot tell what is wrong with this PR. tslint passes locally for me.

src/utils/userDataManager.ts
```diff
@@ -4,30 +4,58 @@ import * as path from 'path';
...
+const restClientDir = 'rest-client';
+const rootPath = path.join(os.homedir(), `.${restClientDir}`);
+
+function getCachePath(): string {
+    if (fs.existsSync(rootPath)) {
+        return rootPath;
+    }
+
+    if (process.env.XDG_CACHE_HOME !== undefined) {
+        return path.join(rootPath, process.env.XDG_CACHE_HOME);
```
Huachao, line 16: Why do we need to join the `rootPath` and `process.env.XDG_CACHE_HOME`?

tristan957: whoops that should be os.homedir().

tristan957: actually is just needs to be process.env.X

src/utils/userDataManager.ts
```diff
@@ -4,30 +4,58 @@ import * as path from 'path';
...
+const rootPath = path.join(os.homedir(), `.${restClientDir}`);
+
+function getCachePath(): string {
+    if (fs.existsSync(rootPath)) {
+        return rootPath;
+    }
+
+    if (process.env.XDG_CACHE_HOME !== undefined) {
+        return path.join(process.env.XDG_CACHE_HOME, restClientDir);
+    }
```
Huachao, line 17: Just a question, if an existing user wants to use the XDG directory, he needs to delete the `.rest-client` folder first. And they also need to export these two environment variables in shell. right?

tristan957: Yes, the user would need to delete it and have the env variables exported. A lot of distributions have these exported by default.

Huachao:
@tristan957 Merged, thanks

tristan957:
Thanks very much for your review

Huachao:
@tristan957 You can try this in the latest version 0.24.0

tristan957:
Perfect I'll let you know how it goes!

PR #591, opened 2020-05-20
README.md: update usage

matyasforian:
Mention that the extension only works if the file has .http extension. This was not clear for me when I installed it, needed to figure out from the screenshots.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ REST Client allows you to send HTTP request and view the response in Visual Stud
     - Fold/Unfold for request block
 
 ## Usage
-In editor, type an HTTP request as simple as below:
+In the editor create a new file, save it with `.http` extension, then type an HTTP request such as:
 ```http
 https://example.com/comments/1
 ```
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ REST Client allows you to send HTTP request and view the response in Visual Stud
     - Fold/Unfold for request block
 
 ## Usage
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,7 +57,7 @@ REST Client allows you to send HTTP request and view the response in Visual Stud
     - Fold/Unfold for request block
 
 ## Usage
-In editor, type an HTTP request as simple as below:
+In the editor create a new file, save it with `.http` extension, then type an HTTP request such as:
```
Huachao, line 60: This is not quite accurate, actually, you don't need to type the request in a file with `.http` extension, just a file of `HTTP` language mode is enough. Even in a plain text file, you can still send HTTP request with Keyboard shortcuts.

matyasforian: This usage section continues with "Once you prepared a request, click the `Send Request` link above the request". This only appears if the extension of the file is `.http*` by default.

I am suggesting this addition because when I used your extension for the first time it took me like 15 mins to figure out why the `send request` button was not showing up. This part is the most important for first-time users, it should contain some super simple instructions how to get things working IMO.
I updated the text, maybe its more accurate now.


>Even in a plain text file, you can still send HTTP request with Keyboard shortcuts.

I tried the `Ctrl+Alt+R` shortcut in a `.css` and in a `.js` file, it did nothing. It worked in a `.http` and `.txt` file.

Huachao:
@matyasforian Merged, thanks 😄

PR #593, opened 2020-05-24
add tree provider

idoprz:
Hi,

This is my first attempt with this extension. I will be happy to fix if I missed few things on the way.

In this PR I added tree view that looks for ".http" files in the workspace and enable to run from the tree.


src/providers/treeViewProvider.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    getTreeItem(element: HttpClientItem): vscode.TreeItem {
+        return element;
+    }
+
+    getChildren(): Thenable<HttpClientItem[]> {
+        return Promise.resolve(this.getDepsInPackageJson());
+    }
+
+    private getDepsInPackageJson(): Thenable<HttpClientItem[]> {
```
Huachao, line 24: A better name, this seems not related to our code

src/extension.ts
```diff
@@ -75,6 +78,17 @@ export async function activate(context: ExtensionContext) {
 
     const diagnosticsProvider = new CustomVariableDiagnosticsProvider();
     context.subscriptions.push(diagnosticsProvider);
+
+    const nodeDependenciesProvider = new HttpTreeProvider();
+    window.registerTreeDataProvider('httpRequest', nodeDependenciesProvider);
+    commands.registerCommand('rest-client.refreshEntry', () => nodeDependenciesProvider.refresh());
```
Huachao, line 84: Can we move this logic into the internal of `HttpTreeProvider `

src/providers/treeViewProvider.ts
```diff
@@ -0,0 +1,63 @@
...
+        return vscode.workspace.findFiles("**/*.http").then((values: vscode.Uri[]) => {
+            return values.map((val: vscode.Uri) => {
```
Huachao, line 26: Personally I prefer the http request tree view corresponds to the active HTTP document in editor. That means the tree view displays all the requests in the current open file, as user change to other http files, the content in tree view will refresh.

idoprz: I don't always need to open the file, so I prefer to just run the request without changing the context of my code editor. So my suggestion is:
1. Make a preference if to just launch the request or to open the editor and launch the request.
2. If there is more then one request in the file, then show the requests as child items in the tree.

Huachao: > 1. Make a preference if to just launch the request or to open the editor and launch the request.

I think we can provide these both, like if click the request URL(or name for named request) in the tree view, it means open the editor, if click the icon in the tree item means sending the request.



> 2. If there is more then one request in the file, then show the requests as child items in the tree.

Personally I think always listing all the requests as the child items are acceptable.

src/providers/treeViewProvider.ts
idoprz: changed it.

src/providers/treeViewProvider.ts
idoprz: i have made the changes according to this discussion.

src/controllers/requestController.ts
```diff
@@ -43,6 +43,10 @@ export class RequestController {
             return;
         }
 
+        return this.runRequest(selectedRequest, document);
+    }
```
Huachao, line 47: Add a blank line between two methods

src/providers/treeViewProvider.ts
```diff
@@ -0,0 +1,82 @@
...
+                        const clients = selectedRequests.map(selectedRequest => {
+                            return HttpClientItem.createRequestItem(document, element.uri, selectedRequest, selectedRequest.name);
+                        });
+                        resolve(clients);
+                    }
+                });
+            });
+        } else {
+            return vscode.workspace.findFiles("**/*.http").then(async (values: vscode.Uri[]) => {
+                return await values.map((uri: vscode.Uri) => {
```
Huachao, line 40: remove the `async` and `await` keywords

src/providers/treeViewProvider.ts
```diff
@@ -0,0 +1,82 @@
...
+    getTreeItem(element: HttpClientItem): vscode.TreeItem {
+        return element;
+    }
+
+    getChildren(element?: HttpClientItem): Thenable<HttpClientItem[]> {
+        return Promise.resolve(this.getHttpClientItems(element));
+    }
+
+    private getHttpClientItems(element): Thenable<HttpClientItem[]> {
+        if (element) {
```
Huachao, line 26: I prefer the design that the `HttpClientItems` holds the range of each request, instead of the parsed ones for the following reasons:

- If you have the range, you don't need to write the method `runRequest`, maybe we can reuse the original `run` method
- If we have range, we can have the capability to click the request in the explorer, and we can go to the actual range of the request

idoprz: I tried that first, but the run method takes the current editor while the requests from the tree doesn't have to have any editor opened since they already has the document. So a change in the run is necessary anyway. I also tried to pass the document to the run but then the code becomes too complex.
What do you think? any suggestions?

PR #602, opened 2020-06-08
Escape quotes for decodeEscapedUnicodeCharacters

Treno1:
Server sends me response as 
```
{
    "text": "\u0022text\u0022"
}
```
My expectation was that decodeEscapedUnicodeCharacters will convert it to
```
{
    "text":"\"text\""
}
```
But result was 
`{"text":""text""}`
with invalid json format errors. 
Curl + jq handles that kind of situation perfectly.

So, small pr to fix that behaviour

### DIFF
--- a/src/utils/httpClient.ts
+++ b/src/utils/httpClient.ts
@@ -254,7 +254,10 @@ export class HttpClient {
     }
 
     private decodeEscapedUnicodeCharacters(body: string): string {
-        return body.replace(/\\u([\d\w]{4})/gi, (_, g) => String.fromCharCode(parseInt(g, 16)));
+        return body.replace(/\\u([\d\w]{4})/gi, (_, g) => {
+            var char = String.fromCharCode(parseInt(g, 16));
+            return char == '\"' ? "\\\"" : char;
+        });
     }
 
     private getRequestCertificate(requestUrl: string): Certificate | null {
--- a/src/utils/httpClient.ts
+++ b/src/utils/httpClient.ts
@@ -255,8 +255,8 @@ export class HttpClient {
 
     private decodeEscapedUnicodeCharacters(body: string): string {
         return body.replace(/\\u([\d\w]{4})/gi, (_, g) => {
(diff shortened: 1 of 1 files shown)


Huachao:
@Treno1 could you please provide the curl response output without jq? A public endpoint is better.

Treno1:
curl -XGET "https://pastebin.com/raw/bZJP0r5i"

src/utils/httpClient.ts
```diff
@@ -254,7 +254,10 @@ export class HttpClient {
     }
 
     private decodeEscapedUnicodeCharacters(body: string): string {
-        return body.replace(/\\u([\d\w]{4})/gi, (_, g) => String.fromCharCode(parseInt(g, 16)));
+        return body.replace(/\\u([\d\w]{4})/gi, (_, g) => {
+            var char = String.fromCharCode(parseInt(g, 16));
```
Huachao, line 258: use `const` instead of `var`

src/utils/httpClient.ts
```diff
@@ -254,7 +254,10 @@ export class HttpClient {
...
+            return char == '\"' ? "\\\"" : char;
```
Huachao, line 259: use `===`

Huachao, line 259: no need to escape the double quote in a single quote, we can simply write like this
```javascript
return char === '"' ? '\\"' : char;
```

Huachao:
> curl -XGET "https://pastebin.com/raw/bZJP0r5i"

Thanks, some minor suggestions.

src/utils/httpClient.ts
Treno1: done

src/utils/httpClient.ts
Treno1: done

Treno1: done

Huachao:
@Treno1 Meged, thanks

Huachao:
@Treno1 you can verify this in the latest version 0.24.0

PR #616, opened 2020-06-18
Add Prompt Variables

theerapatcha:
`Prompt variable` is a variable that will prompt for user input every time user sending a request.

Some variables are meant to be changed all the time such as Session ID or OTP code.
Therefore, I think it is better to let the user input the designated value of that variable instead of changing the source code every time.

### DIFF
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -67,6 +67,6 @@ export const RequestVariableDefinitionRegex: RegExp = RequestVariableDefinitionW
 
 export const NoteCommentRegex = /^\s*(?:#{1,}|\/{2,})\s*@note\s*$/m;
 
-export const PromptCommentRegex = /^\s*(?:#{1,}|\/{2,})\s*@prompt\s*([^\s]+)\s*(.*?)\s*$/m;
+export const PromptCommentRegex = /^\s*(?:#{1,}|\/{2,})\s*@prompt\s+([^\s]+)(?:\s+(.*))?\s*$/m;
 
 export const LineSplitterRegex: RegExp = /\r?\n/g;
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -70,6 +70,6 @@ export const RequestVariableDefinitionWithNameRegexFactory = (name: string, flag
 
 export const RequestVariableDefinitionRegex: RegExp = RequestVariableDefinitionWithNameRegexFactory("\\w+", "m");
 
-export const PromptCommentRegex = /^\s*(?:#{1,}|\/{2,})\s*@prompt\s+([^\s]+)(?:\s+(.*))?\s*$/m;
+export const PromptCommentRegex = /^\s*(?:#{1,}|\/{2,})\s*@prompt\s+([^\s]+)(?:\s+(.*))?\s*$/;
 
 export const LineSplitterRegex: RegExp = /\r?\n/g;
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -67,4 +67,6 @@ export const RequestVariableDefinitionRegex: RegExp = RequestVariableDefinitionW
 
 export const NoteCommentRegex = /^\s*(?:#{1,}|\/{2,})\s*@note\s*$/m;
 
+export const PromptCommentRegex = /^\s*(?:#{1,}|\/{2,})\s*@prompt\s*(.*?)\s*$/m;
+
 export const LineSplitterRegex: RegExp = /\r?\n/g;
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -67,6 +67,6 @@ export const RequestVariableDefinitionRegex: RegExp = RequestVariableDefinitionW
 
 export const NoteCommentRegex = /^\s*(?:#{1,}|\/{2,})\s*@note\s*$/m;
 
-export const PromptCommentRegex = /^\s*(?:#{1,}|\/{2,})\s*@prompt\s*(.*?)\s*$/m;
+export const PromptCommentRegex = /^\s*(?:#{1,}|\/{2,})\s*@prompt\s*([^\s]+)\s*(.*?)\s*$/m;
 
 export const LineSplitterRegex: RegExp = /\r?\n/g;
--- a/src/utils/selector.ts
+++ b/src/utils/selector.ts
@@ -1,12 +1,9 @@
-import { EOL } from "os";
-import { Range, TextEditor, window } from "vscode";
-import * as Constants from "../common/constants";
-import {
-    fromString as ParseReqMetaKey,
-    RequestMetadata,
-} from "../models/requestMetadata";
-import { SelectedRequest } from "../models/SelectedRequest";
-import { VariableProcessor } from "./variableProcessor";
+import { EOL } from 'os';
+import { Range, TextEditor, window } from 'vscode';
+import * as Constants from '../common/constants';
+import { fromString as ParseReqMetaKey, RequestMetadata } from '../models/requestMetadata';
+import { SelectedRequest } from '../models/SelectedRequest';
+import { VariableProcessor } from './variableProcessor';
 
 export interface RequestRangeOptions {
     ignoreCommentLine?: boolean;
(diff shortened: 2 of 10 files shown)


theerapatcha:
@Huachao any chance of this PR getting merged? or is there anything else need to be modified? 

Thanks

Huachao:
@theerapatcha sorry for the late reply, thanks for your contribution, before diving into the implementation, I want to have some discussions with you.

1. The scope of **Prompt Variables** (Request scope or file scope)?
2. The variable definition syntax (Seems related to question 1)
3. The variable reference syntax (I found you add a preceding `$` character before the variable name, actually in my design, this is reserved for system variables only)

theerapatcha:
> @theerapatcha sorry for the late reply, thanks for your contribution, before diving into the implementation, I want to have some discussions with you.
> 
> 1. The scope of **Prompt Variables** (Request scope or file scope)?
> 2. The variable definition syntax (Seems related to question 1)
> 3. The variable reference syntax (I found you add a preceding `$` character before the variable name, actually in my design, this is reserved for system variables only)

1. It is a request scope, our I would say per-request scope just like @note comment tag
2. By adding `// @prompt {varName}` or `# @prompt {varName},{varName2}` just before the request, then, when a user click on `Send Request`, VS Code will prompt for that `varName` values.
3. For now, the prompted input value will replace any reference syntax that matches `$prompt.{varName}`.

To make it align with your design, I can change it to use just plainly `{varName}` syntax? Note that, this prompted variable will be override any variable defined precedingly (just for the prompted request).

Thanks for your reply.

Huachao:
> 1. It is a request scope, our I would say per-request scope just like @note comment tag
> 2. By adding `// @prompt {varName}` or `# @prompt {varName},{varName2}` just before the request, then, when a user click on `Send Request`, VS Code will prompt for that `varName` values.
> 3. For now, the prompted input value will replace any reference syntax that matches `$prompt.{varName}`.

For the first question, does the user requires a file-level prompt variable like the credential information for all APIs?

For the second question, I prefer the syntax that one prompt variable one-line, since I want to add an optional placeholder for a variable description with syntax `# @prompt variableName [variableDescription]` like following:
```http
# @prompt secret My demo service secret
```

For the third question, I think the `{{...}}` seems more aligned with other custom variables

theerapatcha:
> > 1. It is a request scope, our I would say per-request scope just like @note comment tag
> > 2. By adding `// @prompt {varName}` or `# @prompt {varName},{varName2}` just before the request, then, when a user click on `Send Request`, VS Code will prompt for that `varName` values.
> > 3. For now, the prompted input value will replace any reference syntax that matches `$prompt.{varName}`.
> 
> For the first question, does the user requires a file-level prompt variable like the credential information for all APIs?
> 
> For the second question, I prefer the syntax that one prompt variable one-line, since I want to add an optional placeholder for a variable description with syntax `# @prompt variableName [variableDescription]` like following:
> 
> ```httpspec
> # @prompt secret My demo service secret
> ```
> 
> For the third question, I think the `{{...}}` seems more aligned with other custom variables

1. For my use case, i just need a prompt for variable just for a specific request like an OTP from my authenticator app. If there is any need for a file-variable prompt scope, it might be in another PR.
2. That sounds good, will make that change.
3. Got it.

Huachao:
> * For my use case, i just need a prompt for variable just for a specific request like an OTP from my authenticator app. If there is any need for a file-variable prompt scope, it might be in another PR.
> * That sounds good, will make that change.
> * Got it.

The design LGTM, wait for your PR. Thanks for your contribution.

theerapatcha:
I made changes per discussed. You might need to re-review because most of the source code has been changed due to the prompt variable replacement logic which needs to be handled in the `Selector` class.

Btw, I have no idea why LGTM analysis keeps showing failed and I can't find any log that is related.

theerapatcha:
@Huachao resolved the conflicts! does anything need to be changed? Thanks

src/utils/selector.ts
```diff
@@ -170,4 +198,19 @@ export class Selector {
...
+
+    private static async promptForInput(defs: PromptVariableDefinition[]): Promise<Map<string, string> | null> {
+        const promptVariables = new Map<string, string>();
+        for (const { name, description } of defs) {
+            const value = await window.showInputBox({
+                prompt: description ? `${description}` : `Input value for "${name}"`
+            });
+            if (value !== undefined) {
+                promptVariables.set(name, value);
+            } else {
```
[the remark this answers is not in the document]
theerapatcha: Yes. the user is required to input all of the prompt variables defined in a request or else that request will not be firing.

chazmuzz:
Shame this hasn't been merged, was there anything blocking it?

abowen:
Just coming from httpYac (https://httpyac.github.io/), which has this feature, any reason why this can't be merged?

Huachao:
@theerapatcha the only block is that the code has conflicts, I will continue work on this when I am free

theerapatcha:
@Huachao resolved all the conflicts!

Huachao:
@theerapatcha I will review this PR ASAP, it's really a good PR.

src/common/constants.ts
```diff
@@ -70,4 +70,6 @@ export const RequestVariableDefinitionWithNameRegexFactory = (name: string, flag
 
 export const RequestVariableDefinitionRegex: RegExp = RequestVariableDefinitionWithNameRegexFactory("\\w+", "m");
 
+export const PromptCommentRegex = /^\s*(?:#{1,}|\/{2,})\s*@prompt\s+([^\s]+)(?:\s+(.*))?\s*$/m;
```
Huachao, line 73: Do we really need the `m` flag in the regex expression here?

src/utils/selector.ts
```diff
@@ -189,4 +230,29 @@ export class Selector {
...
+            varDefs.push(promptVarDef);
+            metadatas.set(RequestMetadata.Prompt, JSON.stringify(varDefs));
+        }
+    }
+
+    private static async promptForInput(defs: PromptVariableDefinition[]): Promise<Map<string, string> | null> {
+        const promptVariables = new Map<string, string>();
+        for (const { name, description } of defs) {
+            const value = await window.showInputBox({
+                prompt: description ? `${description}` : `Input value for "${name}"`
```
Huachao, line 247: How about always make the prompt text as `Input value for "{name}"`, while setting the `description` as the `placeHolder` when description is available?

Huachao:
@theerapatcha I add some comments, and most of other parts looks really great to me

src/common/constants.ts
theerapatcha: You're right. It is not required.

src/utils/selector.ts
theerapatcha: That's a good idea! :)

theerapatcha:
@Huachao fixed as suggested (as well as one minor misplaced symbol). thanks!

Huachao (approved):

Huachao:
@theerapatcha Merged, thanks for your great contribution

Huachao:
@theerapatcha @chazmuzz @abowen @CaBazaga @jymannob @shuz @unxia now you can try the **Prompt variables** in the latest version 0.25.0

CaBazaga:
Great news, I just tested it and it works perfect. 🎊 

If I may add a suggestion for improvement: Defaults.

Maybe taking any previously globally setted value for the given variable as a default (maybe prefilling the popup field) will work so in case I want to keep the default just hitting enter will work.

Something like
```
@someVar = SomeDefaultValue

### My request.
# @name MyRequest
# @prompt someVar SomeDescription
POST http://...

{
    "varName": "{{someVar}}"
}
```
So having the popup prefilled with the "SomeDefaultValue" allow me just to hit enter to accept it.

So there will be no need to add any extra syntax to achieve it, also if the text on the popup were preselected I can just type to overwrite the default or even just hit "backspace" to erase the popup field to input an "empty" value as it currently does. 

Just a suggestion 😉 

Good work anyway 👍

chazmuzz:
Thank you so much!

chazmuzz:
> So having the popup prefilled with the "SomeDefaultValue" allow me just to hit enter to accept it.

Agree this would be awesome.  Another suggestion in addition to defaults - remembering the previous value entered the last time I was prompted.

PR #620, opened 2020-06-22
Fix broken links in JSON responses

jackbravo:
Links inside the a JSON response are not displayed properly because JSON encodes strings.

This fix displays the original url but the underlying link is using the parsed string as the link `href` property.

This is an example JSON response that has this problem:

```
{
  "authenticateURL": "https://example.com/auth?reg_code=XLMKJDK\u0026mso_id=ATT",
  "expires": 1592867389703
}
```

The parsed link should be `https://example.com/auth?reg_code=XLMKJDK&mso_id=ATT`.

### DIFF
--- a/src/views/httpResponseWebview.ts
+++ b/src/views/httpResponseWebview.ts
@@ -217,7 +217,7 @@ export class HttpResponseWebview extends BaseWebview {
         <div>
             ${this.settings.disableAddingHrefLinkForLargeResponse && response.bodySizeInBytes > this.settings.largeResponseBodySizeLimitInMB * 1024 * 1024
                 ? innerHtml
-                : this.addUrlLinks(innerHtml)}
+                : this.addUrlLinks(innerHtml, contentType)}
             <a id="scroll-to-top" role="button" aria-label="scroll to top" title="Scroll To Top"><span class="icon"></span></a>
         </div>
         <script type="text/javascript" src="${panel.webview.asWebviewUri(this.scriptFilePath)}" nonce="${nonce}" charset="UTF-8"></script>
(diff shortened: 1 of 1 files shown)


src/views/httpResponseWebview.ts
```diff
@@ -351,7 +351,10 @@ ${HttpResponseWebview.formatHeaders(response.headers)}`;
     }
 
     private addUrlLinks(innerHtml: string) {
-        return innerHtml.replace(this.urlRegex, '<a href="$1" target="_blank" rel="noopener noreferrer">$1</a>');
+        return innerHtml.replace(this.urlRegex, match => {
+            const parsedLink = JSON.parse(`"${match}"`);
```
Huachao, line 355: Shall we only do this for JSON response?

jackbravo: Yes! How can we know that?

Huachao: @jackbravo I think you can pass the `contentType` of response from the caller, and in this method just use the `MimeUtility.isJSON` function to check whether the response body is JSON.

jackbravo: done

Huachao:
@jackbravo Merged thanks

Huachao:
@jackbravo you can try the latest version 0.24.2

tangdw:
@Huachao 
JSON 解析出错，响应是这样的：
`{
  test: `src="https://aliyuncs.com/asas/201903/2019032614/file_as.jpg"`
}`

Huachao:
@tangdw the response body you've provided is not JSON actually

> {
> test: `src="[https://aliyuncs.com/asas/201903/2019032614/file_as.jpg"\`](https://aliyuncs.com/asas/201903/2019032614/file_as.jpg%22%5C%60)
> }

jackbravo:
Indeed, that is not json, it needs " surrounding both the key and the value.

On the other hand, the new release is working great for me with links on json. Thanks @Huachao !

Huachao:
@jackbravo @tangdw I got the root cause since @jackbravo will try to JSON parse the links of JSON mime type, here so I think we can simply revert the change and the issue in #676 will be solved. @jackbravo as for your case, I forgot that we already have a setting to handle the escape issue, so you just need to set `rest-client.decodeEscapedUnicodeCharacters": true` in the previous and future versions when I publish a new version

jackbravo:
😮 ! Didn't know about that option. But should work as well.

Issue #500, opened 2020-01-29
excludeHostsForProxy issue


katagaStyle:
Seems that excludeHostsForProxy isn't working wit 0.23.1. 
Rolled back to 0.22.2 and everything works fine.

e14mattc:
Thanks for posting this.. I've been working with this extension over the past day and noticed it can't call out through the proxy I've got set up in VSC (which contains authentication in the form of http://user:pass@proxy:8080). Yet if I roll back to 0.22.2 everything works fine! 

I'd not have realised you could do that if you hadn't have posted!

Looks like the issues are related.

Huachao:
@katagaStyle @e14mattc did you set the setting `http.proxySupport`

e14mattc:
I've turned that off via VSC settings, and updated to 0.23.1 & it's unable to proxy out to the API I'm using. Rolled back to 0.22.2, but kept http.proxySupport = off & it works.

Huachao:
@e14mattc can you confirm that after rollback to 0.22.2 and the `http.proxySupport` is `off`, the auth mechanism passed and the 0.23.1 failed at the **auth** step?

Huachao:
@katagaStyle @e14mattc I'd like to close the issue since I think the `excludeHostsForProxy` setting can work if you set the `http.proxySupport` to `off`

PR #625, opened 2020-06-23
Proxy improvements (enable excludeHostsForProxy )

wtetsu:
Hello, I love this extension and use everyday!

I noticed two proxy issues, so I tried fixed them.

I thought to fix the two at the same is convenient.
so this PR deal with the both two.

# Summary

* (1) `rest-client.excludeHostsForProxy` doesn't work
* (2)`http.proxy` could affect the whole VSCode
  * Hopefully, I want to set a proxy that affect only this extension (e.g. `rest-client.proxy`)

# Reproducible Environment

* 0.24.1 and master(ceab9394dcec576ea38bcc83d53b086317d87d95)
* OS: OSX / Windows 10
* VSCode: 1.46.1

# (1) `excludeHostsForProxy` doesn't work

Maybe already reported #500

In this case, `excludeHostsForProxy` doesn't work.
```
(settings)
"http.proxy": "http://aaa.bbb.com:8888",
"rest-client.excludeHostsForProxy": ["httpbin.org"]

----
GET https://httpbin.org/get

```

I debugged and finally discovered, this "set proxy" block doesn't work.

https://github.com/Huachao/vscode-restclient/blob/master/src/utils/httpClient.ts#L176-L191

Even if you delete the 15 lines, `http.proxy` works, and `excludeHostsForProxy` doesn't work.

I changed the code to use tunnel that follows Got's example.

https://github.com/sindresorhus/got#proxies

# (2)`http.proxy` could affect the whole VSCode

Hopefully, if there is a configuration that affects only this extension, that's really useful.

Because `http.proxy` is a VSCode's configuration that could affect everywhere.

I tried adding `rest-client.proxy`.

# Examples

## Current version (0.24.1)

`excludeHostsForProxy` doesn't work.

```
{
  "http.proxy": "http://aaa.bbb.com:8888",
  "rest-client.excludeHostsForProxy": [
    "httpbin.org"
  ]
}
```

## This PR version

`excludeHostsForProxy` works.
When calling httpbin.org, the proxy is not used.

```
{
  "http.proxy": "http://aaa.bbb.com:8888",
  "rest-client.excludeHostsForProxy": [
    "httpbin.org"
  ]
}

```

This also works.

```
{
  "rest-client.proxy": "http://aaa.bbb.com:8888"
  "rest-client.excludeHostsForProxy": [
    "httpbin.org"
  ]
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -371,6 +371,12 @@
           "scope": "resource",
           "description": "Send out anonymous usage data"
         },
+        "rest-client.proxy": {
+          "type": "string",
+          "default": null,
+          "scope": "resource",
+          "description": "The proxy setting to use"
+        },
         "rest-client.excludeHostsForProxy": {
           "type": "array",
           "default": [],
@@ -666,6 +672,7 @@
     "pretty-data": "^0.40.0",
     "tough-cookie": "^3.0.1",
     "tough-cookie-file-store-bugfix": "Huachao/tough-cookie-file-store-bugfix",
+    "tunnel": "0.0.6",
     "uuid": "^3.3.2",
     "xmldom": "^0.1.27",
     "xpath": "^0.0.27",
--- a/src/models/configurationSettings.ts
+++ b/src/models/configurationSettings.ts
@@ -156,7 +156,7 @@ export class RestClientSettings implements IRestClientSettings {
         languages.setLanguageConfiguration('http', { brackets: this.addRequestBodyLineIndentationAroundBrackets ? this.brackets : [] });
 
         const httpSettings = workspace.getConfiguration("http");
-        this.proxy = httpSettings.get<string>('proxy');
+        this.proxy = restClientSettings.get<string>("proxy") ?? httpSettings.get<string>('proxy');
         this.proxyStrictSSL = httpSettings.get<boolean>('proxyStrictSSL', false);
     }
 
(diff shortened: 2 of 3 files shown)


Huachao:
@wtetsu thanks for your contribution, but I stlll have two questions to discuss with you:

1. Can we reuse the `http.proxy` setting since I think users don't want to create another proxy setting
2. I found the proxy network behavior is also controlled by `"http.proxySupport"` setting whe the `http.proxy` is set

wtetsu:
@Huachao 

Thank you for your comment.

1. Yes, of couse I can delete the new `rest-client.proxy` if you don't want it. But it seems the whole this PR should be abolished due to 2.

2. I didn't know `proxySupport` so I tried it. In conclusion, the below worked perfectly. The both `http.proxy` and `excludeHostsForProxy` worked perfectly without this PR's modification.

```
  "http.proxy": "http://aaa.bbb.com:8888",
  "http.proxySupport": "off",
  "rest-client.excludeHostsForProxy": [
    "httpbin.org"
  ]
```

So, I'm going to close this PR.  Sorry for bothering you. But I hope this thread helps users who faced "excludeHostsForProxy doesn't work" issue.

Huachao:
@wtetsu since VSCode added the proxy support for extensions, we should set the `http.proxySupport` to `off` as you did if you expect the "rest-client.excludeHostsForProxy" to work. I can add some notes in the README.md.

PR #626, opened 2020-06-23
Fix some typo

wtetsu:
Hello, when I was reading the source code, I noticed some trivial typo.
Just in case I create a PR that fixes them.


Huachao:
@wtetsu Merged, thanks

PR #632, opened 2020-07-03
processing variables from external file

ricmrodrigues:
this is part of an issue I've raised, maybe missing something but I tested and it seemed to work :)

### DIFF
--- a/src/providers/documentLinkProvider.ts
+++ b/src/providers/documentLinkProvider.ts
@@ -7,7 +7,7 @@ import { getWorkspaceRootPath } from '../utils/workspaceUtility';
 
 export class RequestBodyDocumentLinkProvider implements DocumentLinkProvider {
 
-    private _linkPattern = /^(\<\s+)(.+)(\s*)$/g;
+    private _linkPattern = /^(\<(?:@?(?:\w+)?)?\s+)(.+)(\s*)$/g;
 
     public provideDocumentLinks(document: TextDocument, _token: CancellationToken): DocumentLink[] {
         const results: DocumentLink[] = [];
--- a/src/providers/documentLinkProvider.ts
+++ b/src/providers/documentLinkProvider.ts
@@ -7,7 +7,7 @@ import { getWorkspaceRootPath } from '../utils/workspaceUtility';
 
 export class RequestBodyDocumentLinkProvider implements DocumentLinkProvider {
 
-    private _linkPattern = /^(\<(?:@?(?:\w+)?)?\s+)(.+)(\s*)$/g;
+    private _linkPattern = /^(\<(?:@(?:\w+)?)?\s+)(.+)(\s*)$/g;
 
     public provideDocumentLinks(document: TextDocument, _token: CancellationToken): DocumentLink[] {
         const results: DocumentLink[] = [];
--- a/src/utils/httpRequestParser.ts
+++ b/src/utils/httpRequestParser.ts
@@ -187,8 +187,7 @@ export class HttpRequestParser implements RequestParser {
                                 const fileContent = buffer.toString(groupsValues.encoding || this.defaultFileEncoding);
                                 const resolvedContent = await VariableProcessor.processRawRequest(fileContent);
                                 combinedStream.append(resolvedContent);
-                            }
-                            else {
+                            } else {
                                 combinedStream.append(fs.createReadStream(fileAbsolutePath));
                             }
                         } else {
--- a/src/utils/httpRequestParser.ts
+++ b/src/utils/httpRequestParser.ts
@@ -8,6 +8,7 @@ import { RequestParser } from '../models/requestParser';
 import { MimeUtility } from './mimeUtility';
 import { getContentType, getHeader, removeHeader } from './misc';
 import { parseRequestHeaders, resolveRequestBodyPath } from './requestParserUtil';
+import { VariableProcessor } from "./variableProcessor";
 
 const CombinedStream = require('combined-stream');
 const encodeurl = require('encodeurl');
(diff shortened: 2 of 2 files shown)


src/utils/httpRequestParser.ts
```diff
@@ -179,7 +180,10 @@ export class HttpRequestParser implements RequestParser {
                         const inputFilePath = groups[1];
                         const fileAbsolutePath = await resolveRequestBodyPath(inputFilePath);
                         if (fileAbsolutePath) {
-                            combinedStream.append(fs.createReadStream(fileAbsolutePath));
+                            const buffer = await fs.readFile(fileAbsolutePath);
+                            const fileContent = buffer.toString('utf8');
```
Huachao, line 184: Here you simply convert the buffer to UTF-8 string, if the buffer encoding is not `utf8`, this would make the final request body incorrect

src/utils/httpRequestParser.ts
```diff
@@ -179,7 +180,10 @@ export class HttpRequestParser implements RequestParser {
...
+                            const resolvedContent = await VariableProcessor.processRawRequest(fileContent);
+                            combinedStream.append(resolvedContent);
```
Huachao, line 186: If we really want to support this feature, I think we should add a switch(setting) to turn on/off this behavior that parsing the variables in the included body file.

src/utils/httpRequestParser.ts
ricmrodrigues: what's the best way of doing it?

src/utils/httpRequestParser.ts
ricmrodrigues: overall setting? that seems a bit excessive, why not a request level setting?

src/utils/httpRequestParser.ts
Huachao: I found that VSCode interally use the library `jschardet` to guess textual file content encoding (https://github.com/microsoft/vscode/blob/ce1462776f7ada735ba82ffec28ea774e617a927/src/vs/workbench/services/textfile/common/encoding.ts#L234). Together with the following advice, if we add the request level setting for this feature, maybe we can also support providing an optional argument `encoding`, if omit, we will guess the encoding for him/her. By default the setting is turned off.

src/utils/httpRequestParser.ts
ricmrodrigues: like instead of `< filename` we could do `<@ filename` to imply it contains variables needing resolving ? What do you think

src/utils/httpRequestParser.ts
ricmrodrigues: think we can default to utf8 and provide a way to override the default to keep it simple

ricmrodrigues: added those changes, can you check?

src/utils/httpRequestParser.ts
ricmrodrigues: added this with an optional encoding, seems like a nice option, brings in the default with an option to override and only processes variables in files if we signal it that way with the `@`

src/utils/httpRequestParser.ts
```diff
@@ -175,11 +177,20 @@ export class HttpRequestParser implements RequestParser {
...
                         const fileAbsolutePath = await resolveRequestBodyPath(inputFilePath);
                         if (fileAbsolutePath) {
-                            combinedStream.append(fs.createReadStream(fileAbsolutePath));
+                            if (groupsValues.processVariables) {
+                                const buffer = await fs.readFile(fileAbsolutePath);
+                                const fileContent = buffer.toString(groupsValues.encoding || this.defaultFileEncoding);
+                                const resolvedContent = await VariableProcessor.processRawRequest(fileContent);
+                                combinedStream.append(resolvedContent);
+                            }
+                            else {
```
Huachao, line 191: Please fix the tslint error

src/utils/httpRequestParser.ts
Huachao: nice suggestion

src/utils/httpRequestParser.ts
ricmrodrigues: fixed

src/utils/httpRequestParser.ts
```diff
@@ -21,7 +22,8 @@ enum ParseState {
 export class HttpRequestParser implements RequestParser {
     private readonly defaultMethod = 'GET';
     private readonly queryStringLinePrefix = /^\s*[&\?]/;
-    private readonly inputFileSyntax = /^<\s+(.+?)\s*$/;
```
Huachao, line 24: The `documentLInkProvider.ts` also uses the same regex to identify the file link, you also need to update the regex there since the Click link feature is broken when appendig `@` after `<`

ricmrodrigues: very good point, noticed the underscore stopped working after my change!

src/providers/documentLinkProvider.ts
```diff
@@ -7,7 +7,7 @@ import { getWorkspaceRootPath } from '../utils/workspaceUtility';
 
 export class RequestBodyDocumentLinkProvider implements DocumentLinkProvider {
 
-    private _linkPattern = /^(\<\s+)(.+)(\s*)$/g;
+    private _linkPattern = /^(\<(?:@?(?:\w+)?)?\s+)(.+)(\s*)$/g;
```
Huachao, line 10: It seems in your regex, the encoding is supported with the `<` syntax, in my option, it's only supported with the `<@` syntax

ricmrodrigues: sorted

Huachao:
@ricmrodrigues Merged, thanks

jackbravo:
What was missing was documenting how this feature works in the README file.

ricmrodrigues:
> What was missing was documenting how this feature works in the README file.

https://github.com/Huachao/vscode-restclient/pull/637

Huachao:
@ricmrodrigues you can try the latest version 0.24.2

PR #654, opened 2020-07-31
Add support for text/javascript as JSON

mingue:


mingue:
@Huachao any chance on getting this merged? it's quite a straightforward change that will allow us to use this as test suite at work. Thanks!

Huachao:
@mingue I think simply treat javascript MIME type as JSON seems inappropriate

mingue:
what are the implications of it? as far as I can see it's only blocking me to use that response on other requests

Huachao:
@mingue Personally I prefer treating `text/javascript` as a different mime type as `application/json`, so I mean in the `mimeUtility.ts` they should locate in two different methods. However, you can try to treat them in the same way when they use, like when you want to format the response in the JSON way, you can try to check if the response mime type is JSON or JavaScript.

mingue:
Hi @Huachao I followed your recommendations to get this patched. Let me know is this is good to merge.

mingue:
@Huachao any chance on taking a look into this?

jchirschy:
Hello @mingue,

Would it be preferable to use `t.includes("json")` over `t.endsWith("+json")`. It is more compatible with `Content-Type` like `application/x-amz-json-1.1`

src/utils/mimeUtility.ts
```diff
@@ -77,7 +77,7 @@ export class MimeUtility {
         return essence === 'application/xml' || essence === 'text/xml' || subtype.endsWith('+xml');
     }
 
-   public static isHtml(contentTypeString: string | undefined): boolean {
+    public static isHtml(contentTypeString: string | undefined): boolean {
```
Huachao, line 80: add back the leading space

src/utils/mimeUtility.ts
```diff
@@ -90,7 +90,8 @@ export class MimeUtility {
             return false;
         }
 
-        return this.parse(contentTypeString).essence === 'application/javascript';
+        const essence = this.parse(contentTypeString).essence;
+        return essence === 'application/javascript' || essence.includes("text/javascript");
```
Huachao, line 94: Why we use `includes` here

src/utils/requestVariableCacheValueProcessor.ts
```diff
@@ -58,7 +58,7 @@ export class RequestVariableCacheValueProcessor {
             }
 
             const contentTypeHeader = getContentType(headers);
-            if (MimeUtility.isJSON(contentTypeHeader)) {
+            if (MimeUtility.isJSON(contentTypeHeader) || (MimeUtility.isJavaScript && isJSONString(body as string))) {
```
Huachao, line 61: `isJavaScript(contentTypeHeader)`

Huachao (changes requested):

src/utils/mimeUtility.ts
mingue: They are actually aligned now

src/utils/mimeUtility.ts
mingue: as it might contain some other text i.e: "Content-Type: text/javascript; charset=utf-8"

src/utils/requestVariableCacheValueProcessor.ts
mingue: fixed

mingue:
> Hello @mingue,
> 
> Would it be preferable to use `t.includes("json")` over `t.endsWith("+json")`. It is more compatible with `Content-Type` like `application/x-amz-json-1.1`

The change is to add support for "text/javascript" as valid json for a response to be able to use it later on following requests

src/utils/mimeUtility.ts
Huachao: no need since after the parsing logic, the essence would not include the charset informationion

Huachao (changes requested):

src/utils/mimeUtility.ts
mingue: Fixed!

Huachao (approved):

Huachao:
@mingue Merged, thanks.

PR #664, opened 2020-08-17
Add codelens provider to show current environment

lee-baker:
This is my first time offering a pull request, I have tried to maintain the same styles as the project.
I have made the view dependent on the same setting as sending the request via codelens. Could add a different setting for this if you think that would be better.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -12,6 +12,7 @@ import { EnvironmentOrFileVariableHoverProvider } from './providers/environmentO
 import { FileVariableDefinitionProvider } from './providers/fileVariableDefinitionProvider';
 import { FileVariableReferenceProvider } from './providers/fileVariableReferenceProvider';
 import { FileVariableReferencesCodeLensProvider } from './providers/fileVariableReferencesCodeLensProvider';
+import { EnvironmentCodeLensProvider } from './providers/EnvironmentCodeLensProvider';
 import { HttpCodeLensProvider } from './providers/httpCodeLensProvider';
 import { HttpCompletionItemProvider } from './providers/httpCompletionItemProvider';
 import { HttpDocumentSymbolProvider } from './providers/httpDocumentSymbolProvider';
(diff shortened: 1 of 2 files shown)


Huachao:
@lee-baker actually current environment is already displayed at the right bottom of the status bar. The code exists in `src/utils/environmentStatusBarEntry.ts`. And I don't think every request needs this code lens since the switch environment operation is not of the request level, but the VS Code instance level, so I think only one place to switch the environment is enough.

lee-baker:
I am suggesting this more as a way to verify which environment is the current one, that the request variables will come from, than switching environment

Huachao:
@lee-baker as you can see in the right bottom of the status bar, you can already see the environment name and click it will pop-up the switch environment dialog.

lee-baker:
You are very correct, and I know they already exist, however when using this extension I have too many times made a request to the wrong environment because I don't always check the status bar (even though I know I should). I thought it might be an issue others also face and putting the environment next to the button to make the request means its much harder to make this mistake. I understand if you don't think this belongs in the main project, but I know I would find it very helpful.

Huachao:
@lee-baker, in my opinion, it's not a must-have feature, and we can leave the PR open and wait for the community feedback

iainwlg:
Im for it. I have been got by this multiple times. I know the indicator is there, but it doesn't exactly stand out.

idonothaveacat:
I'm all for this. When you're working across multiple environments is easy to forget which one you are on. It's not obvious in the bottom corner of the screen. I consider this a major UX improvement.

edwardpointon:
I accidentally clobbered an environments config, which took me 4 hours to get backups and recover.. because i didn't check the tiny status bar.... please merge this in!!!

lee-baker:
@Huachao There has been some movement from the community here. Is this something you would consider adding now?

Another option would be to expose the environment and change environment command via an api exposed from your activate function. I could then create a different extension that depends on yours but gives this extra functionality.

What do you think?

PR #691, opened 2020-10-01
Search for .env files on parent folders

### LINKED ISSUE #690: Support .env file on the root of the project
This extension do not support `.env` files that are not on the same folder as the `.http` file

Imagine this project structure:
```
.env
src/
  users/
    create/
      handler.ts
      doc.http
    list/
      handler.ts
      doc.http
```
On each subfolders `doc.http` file, it would be great to use variables that are defined on the project root `.env` file.

joaogsleite:
This extension just supports `.env` files that are on the same folder of the `.http` file.

With this change, the extension will search for `.env` files on parent folders until it finds one.

closes #690


Huachao (approved):

Huachao:
@joaogsleite nice patch, merged, thanks

PR #738, opened 2020-12-24
Update README.md

pauldendulk:
Under usage start with 'In Visual Studio Code create a file with an .http extension'.

Why: It was not clear to me what 'In the editor' meant. I only understood this after reading the 'by default .http files are like this' remark later on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ REST Client allows you to send HTTP request and view the response in Visual Stud
     - Fold/Unfold for request block
 
 ## Usage
-In editor, type an HTTP request as simple as below:
+In Visual Studio Code create a file with an `.http` extension and enter an HTTP request as simple as below:
 ```http
 https://example.com/comments/1
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,7 +72,8 @@ content-type: application/json
...
+Once you prepared a request, click the `Send Request` link above the request (this will appear if the file's language mode is `HTTP`, by default `.http` files are like this), or use shortcut `Ctrl+Alt+R`(`Cmd+Alt+R` for macOS), or right-click in the editor and then select `Send Request` in the menu, or press `F1` and then select/type `Rest Client: Send Request`, the response will be previewed in a separate __webview__ panel of Visual Studio
```
Huachao, line 75: **Visual Studio Code** not **Visual Studio**？

pauldendulk: Visual Studio Code and Visual Studio are two different editors. Your plugin works for VS Code and I assumed not for VS. Is this correct?

Huachao: Yes, so I think you miss the word `Code` in the sentence
> the response will be previewed in a separate __webview__ panel of Visual Studio.

Huachao (changes requested):

Issue #737, opened 2020-12-23
Default headers per each file


mafshin:
I know there is `rest-client.defaultHeaders` setting available for defining default headers but that setting is part of settings on a user machine and is global for all requests. 

How can we achieve something similar per each .http file? So that when the file is shipped to someone else it has all of its dependencies (default-headers).

It's quite useful when you have multiple services and one .http file per each service with different headers/

**ServiceA.http**
```
 defaultHeaders: {
  Customheader1: val1
}
```

**ServiceB.http**
```
 defaultHeaders: {
 Customerheader2: val2
}
```

Huachao:
@mafshin one way to achieve this is to define the file variables and reference them in each request

mafshin:
I already use variables for request headers but when each request has multiple default headers and you have tens of request per each file then most of the lines in the file are repeated just for headers. So having this feature will greatly improve the user experience of REST Client.

PR #762, opened 2021-02-01
Added FileHeader support with @@ syntax

mafshin:
I added support for specifying header variables with **@@** prefix (#737 ).

The file headers are processed based on the position of http request so that any file header above the request is automatically added to the list of request headers.

Also variable expansion in file headers are supported.

Sample usage:

```
@url  = https://jsonplaceholder.typicode.com/todos/1
@@Content-Type: application/json

### Request 1 with Content-Type header
GET {{url}}

### new header
@token = myAuthToken
@@Authorization: Bearer {{token}}


### Request 2 with Content-Type, Auhtorization and Accept-Encoding headers
GET {{url}}
Accept-Encoding: gzip
````

### DIFF
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -61,7 +61,7 @@ export const AzureClouds: { [key: string]: { aad: string, arm: string, armAudien
 
 export const CommentIdentifiersRegex: RegExp = /^\s*(#|\/{2})/;
 
-export const FileVariableDefinitionRegex: RegExp = /^\s*@([^\s=]+)\s*=\s*(.*?)\s*$/;
+export const FileVariableDefinitionRegex: RegExp = /^\s*@([^\s=@]+)\s*=\s*(.*?)\s*$/;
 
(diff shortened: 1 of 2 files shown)


mafshin:
@Huachao this feature makes the .http request file much smaller and more readable specially when you have several default headers for a service. 

The main other benefit of having these default variables as part of request file instead of user settings is that you can have them versioned in source control and share them with team as API is evolving.

guillaume86:
I'm interested in this, any news regarding this PR?
Repeating the same headers every request lowers the signal/noise ratio a lot in http files.
(a special baseURL variable would also be nice but it's out of scope of this PR)

mafshin:
@Huachao no intention to merge this pull request or a similar implementaion? 

Currently the request in .http files occupy multiple lines so this feature can greatly minimze the number of lines and makes finding the request much much easier.

Specially since there can be a .http file per Api and there exist common headers among each Api requests, then this feature is a game-changer.

PR #794, opened 2021-03-17
#698 add :/ in scope regex to match

### LINKED ISSUE #698: AADV2TokenProvider only works for special case of MS Graph Tokens
- REST Client Version: 0.24.3
- VSCode Version: 1.50.1
- OS Version: Win 10 2004

Steps to Reproduce:

1. Make HTTP request using `$aadV2Token` using full scope format:
```
GET https://graph.microsoft.com/v1.0/me
Authorization: {{$aadV2Token new scopes:https://graph.microsoft.com/User.Read}}
```
2. AuthParameters regex fails to parse scopes, and request fails:
```
Auth call failed. AADSTS70011: The provided value for the input parameter 'scope' is not valid. The scope 'https openid profile' does not exist.
Trace ID: 983bf6d2-935e-473d-908b-1852c4d74d00
Correlation ID: eb41592f-31cd-4da2-8d40-468761cdb892
Timestamp: 2020-10-18 18:03:45Z
```

#649 is a tremendous contribution by @darrelmiller, and I was really excited to see this in place!

However, the Regex used to parse the token declaration (https://github.com/Huachao/vscode-restclient/blob/f377477ad5802da7575d77047a12d3eba633d22f/src/utils/aadV2TokenProvider.ts#L146), only works for the special case of MS Graph scopes (using a shortened format), and fails when using the generic scope format supported by the MS Identity endpoint. That limits the use of this token provider unnecessarily.

In order to provide a comprehensive migration path from the v1.0 legacy endpoint to the new Microsoft Identity (v2.0) endpoint, the non-standard `resource` parameter from AAD V1 has been ported over into V2 scopes, see here: https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-v1-app-scopes

Any scopes declared are composed of the Resource Uri followed by the Application-specific scope name. Hence, MS Graph scopes on the V2 endpoint look like this: `https://graph.microsoft.com/User.Read`, and a valid Power BI API scope would look like this: `https://analysis.windows.net/powerbi/api/Tenant.Read.All`, for example.

It just happens to be the case that for MS Graph specifically the "https://graph.microsoft.com/" resource uri can be omitted, i.e. "User.Read" is sufficient in the above example.

Not supporting the generic case excludes all other Azure resources from being usable here.

Please adjust the regex accordingly.

abdulkareemnalband:
{{$aadV2Token scopes:https://graph.windows.net/openid clientId:<clientId>  }}
fixes #698

### DIFF
--- a/src/utils/aadV2TokenProvider.ts
+++ b/src/utils/aadV2TokenProvider.ts
@@ -143,7 +143,7 @@ export class AadV2TokenProvider {
 */
 class AuthParameters {
 
-    private readonly aadV2TokenRegex: RegExp = new RegExp(`\\s*\\${Constants.AzureActiveDirectoryV2TokenVariableName}(\\s+(${Constants.AzureActiveDirectoryForceNewOption}))?(\\s+(appOnly))?(\\s+scopes:([\\w,./:]+))?(\\s+tenantId:([^\\.]+\\.[^\\}\\s]+|[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}))?(\\s+clientId:([^\\.]+\\.[^\\}\\s]+|[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}))?\\s*`);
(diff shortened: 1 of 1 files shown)


pschaeflein:
When calling custom apis that are protected by Azure AD, and if using the default values provided in the Azure Portal, you end up with a scope similar to `api://<guid>/scope.name`. (Or, if following some of the ASP.Net samples, the `api://<guid>/scope_name`).

In this scenario, `\S` is a better alternative than simply adding a semi-colon (`:`).

drub0y:
Just hitting this myself now which really puts a stop to everything I was hoping to do with `.rest` files for my API which is secured with custom scopes that necessitate these currently invalid characters. Is there anything we can do to help get this PR pushed through?

While I agree with @pschaeflein change, the current change would at least technically address the issues for a lot of people. We can always do a follow up PR to use `\S` later.

pschaeflein:
I've resorted to running a private install based on a fork. I asked about additional maintainers, but no "official" response...

https://github.com/Huachao/vscode-restclient/discussions/826

Huachao (approved):

Huachao:
@abdulkareemnalband @pschaeflein @drub0y thanks for your PR and suggestions, and I have changed to use `\S` instead, this fix will be published in the next release.

dracan:
@Huachao - Do you know when this will be released? I've just hit the same issue too, and it's quite a blocker.

Ps. awesome work on this extension - it's amazing! :)

PR #795, opened 2021-03-17
Update azure ad token caching

abdulkareemnalband:
1. include scopes in cache key
       different network request may need token with different scopes , so include scopes in cache key
2. include original requested scopes also in cache scope
       Sometime token is missing requested scopes.
       When next request is sent with same scope application is asking for re authentication.
       Including original scopes in saved scopes fixes that

### DIFF
--- a/src/utils/aadV2TokenProvider.ts
+++ b/src/utils/aadV2TokenProvider.ts
@@ -42,7 +42,7 @@ export class AadV2TokenProvider {
         }
     }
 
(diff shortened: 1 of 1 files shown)


Huachao:
@abdulkareemnalband Merged, thanks

Huachao:
@abdulkareemnalband this has been published in the latest version 0.24.5, thanks so much

PR #808, opened 2021-04-06
Support cUrl SSL client certificate options

cplussharp:
extend cUrl parser to parse:
* -E or --cert
* --key
* --pass
* --cacert

extend httpRequest to have https options.

prefere https options of the request
over the https options from the settings


cplussharp:
This should now cover every places: httpParser, curlParser, syntax highlighting, variable support, history, curl command generation and output of the parameters in response view. The only thing not working is the CodeSnipped generator, because "httpsnipped" does not support client certificates :-(

PR #852, opened 2021-05-21
Fix issue with URL links picking up trailing quotes (fixes #816)

### LINKED ISSUE #816: hyperlink address in response includes trailing double quote (version 0.24.5)
My JSON response contains download links for files like
{
  ...
  "url": "https://path/to/file.ext"
}

RestClient v0.24.5 creates a hyperlink from it with the following target address:
  https://path/to/file.ext"
When I click the hyperlink, it won't resolve because of the trailing double quote added by RestClient.

In v0.24.4 this worked fine. RestClient created target address as expected:
  https://path/to/file.ext

adibradfield:
An update to highlight.js means that it has now started escaping quotes, which has broken the regex that is used to find and replace urls. This PR adds additional logic to remove the encoded characters from the link that gets generated

### DIFF
--- a/src/views/httpResponseWebview.ts
+++ b/src/views/httpResponseWebview.ts
@@ -353,7 +353,22 @@ ${HttpResponseWebview.formatHeaders(response.headers)}`;
     }
 
     private addUrlLinks(innerHtml: string) {
-        return innerHtml.replace(this.urlRegex, '<a href="$1" target="_blank" rel="noopener noreferrer">$1</a>');
+        return innerHtml.replace(this.urlRegex, (match: string): string => {
+            const encodedEndCharacters = ["&lt;","&gt;","&quot;","&apos;"];
+            let urlEndPosition = match.length;
+
+            encodedEndCharacters.forEach((char) => {
+                var index = match.indexOf(char);
+                if(index > -1 && index < urlEndPosition){
+                    urlEndPosition = index;
+                }
+            });
+
+            var url = match.substr(0, urlEndPosition);
+            var extraCharacters = match.substr(urlEndPosition);
+
+            return '<a href="' + url + '" target="_blank" rel="noopener noreferrer">' + url + '</a>' + extraCharacters;
+        });
     }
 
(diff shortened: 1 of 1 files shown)


src/views/httpResponseWebview.ts
```diff
@@ -353,7 +353,22 @@ ${HttpResponseWebview.formatHeaders(response.headers)}`;
     }
 
     private addUrlLinks(innerHtml: string) {
-        return innerHtml.replace(this.urlRegex, '<a href="$1" target="_blank" rel="noopener noreferrer">$1</a>');
+        return innerHtml.replace(this.urlRegex, (match: string): string => {
+            const encodedEndCharacters = ["&lt;","&gt;","&quot;","&apos;"];
+            let urlEndPosition = match.length;
+
+            encodedEndCharacters.forEach((char) => {
+                var index = match.indexOf(char);
```
Huachao, line 361: use `let` or `const` over `var`

src/views/httpResponseWebview.ts
```diff
@@ -353,7 +353,22 @@ ${HttpResponseWebview.formatHeaders(response.headers)}`;
...
+                if(index > -1 && index < urlEndPosition){
+                    urlEndPosition = index;
```
Huachao, line 363: is it just simply
```js
if (index !== -1)
```

src/views/httpResponseWebview.ts
```diff
@@ -353,7 +353,22 @@ ${HttpResponseWebview.formatHeaders(response.headers)}`;
     }
 
     private addUrlLinks(innerHtml: string) {
-        return innerHtml.replace(this.urlRegex, '<a href="$1" target="_blank" rel="noopener noreferrer">$1</a>');
```
Huachao, line 356: Can we do this directly in the `urlRegex`?

Huachao (changes requested):

Huachao (approved):

Huachao:
@adibradfield Merged, thanks

Issue #924, opened 2021-11-14
"Save Response Body" to get output filename from Content-Disposition (if available)


mbronk:
Currently, the **Save Response Body** option does auto-pick an extension based on Content-Type
Ref:
https://github.com/Huachao/vscode-restclient/blob/98fcad29c7d9416133df23129730253045662f29/src/views/httpResponseWebview.ts#L147

It would be nice, if it also picked up the server-provided name, provided through **Content-Disposition** (for "downloads").
E.g. save to _test.zip_, for a `Content-Disposition: attachment; filename="test.zip"` (similar to what curl's _--remote-name --remote-header-name_ does)

Huachao:
@mbronk this has been published in the latest version 0.24.6 and you can have a try

mbronk:
@Huachao - Just gave 0.24.6 a spin - works nicely! Thanks for integrating this enhancement.

PR #925, opened 2021-11-14
SaveBody now uses filename from content-disposition

mbronk:
Closes: #924 
(naive implementation, mostly as a proof of concept)

New behavior is toggle-able via useContentDispositionFilename setting (default true).
Only applies to "attachment" and "inline" disposition types (even though inline+filename response is not specifically following the syntax, it is often encountered in practice).

Note (on security): The extension does NOT sanitize the output filename and uses server-provided - equvalent of curl's --remote-header-name option).


Huachao (approved):

Huachao:
@mbronk Thanks for your contribution

PR #933, opened 2021-11-29
Add Markdown support to run requests with code lenses

acuarica:
This PR allows you to run your requests directly from a `markdown` file. It displays a code lens `Send Request` when it detects a new code block using `http` language.

rest-client-md-support

This has been previously discussed in https://github.com/Huachao/vscode-restclient/issues/53#issuecomment-265009922. There is no new shortcuts, only displaying code lenses. Moreover, if you select `Send Request` from the command palette, it will run the code block where the active line is in. I hope this does not cause any conflict.

The goal behind this PR is to make Markdown to be live documentation, in the spirit of org-mode. Your extension is the perfect tool to implement this feature. Thanks for the good work!
 
Looking forward to hear your comments.

PR #938, opened 2021-12-08
Support for AsciiDoc fenced code blocks

wouteraj:
First of all, thanks for the great extension. My team and I have been using it extensively in the last couple of months, and we consider this to be an essential part of our workflow. One useful addition would be to allow requests to be sent from `asciidoc`-files. This would enable tighter integration of writing documentation in our workflow.

I've created this PR for exactly this reason and based it on the very similar PR #933.

Happy to have your feedback on this!


woutermont:
More general, it would also be nice to let the user enable this for any language (e.g. using a custom delimiter), or classes of languages (plain text styles, markdown dialects etc.).

PR #1009, opened 2022-05-22
Add curl argument: --data-urlencode

mikecanann:
Add a command line argument for curl, called: --data-urlencode

### DIFF
--- a/src/utils/curlRequestParser.ts
+++ b/src/utils/curlRequestParser.ts
@@ -52,7 +52,7 @@ export class CurlRequestParser implements RequestParser {
         }
 
         // parse body
-        let body = parsedArguments.d || parsedArguments.data || parsedArguments['data-ascii'] || parsedArguments['data-binary'] || parsedArguments['data-raw'];
+        let body = parsedArguments.d || parsedArguments.data || parsedArguments['data-ascii'] || parsedArguments['data-binary'] || parsedArguments['data-raw'] || parsedArguments['data-urlencode'];
         if (Array.isArray(body)) {
             body = body.join('&');
         } else  {


src/utils/curlRequestParser.ts
```diff
@@ -52,7 +52,7 @@ export class CurlRequestParser implements RequestParser {
         }
 
         // parse body
-        let body = parsedArguments.d || parsedArguments.data || parsedArguments['data-ascii'] || parsedArguments['data-binary'] || parsedArguments['data-raw'];
+        let body = parsedArguments.d || parsedArguments.data || parsedArguments['data-ascii'] || parsedArguments['data-binary'] || parsedArguments['data-raw'] || parsedArguments['data-urlencode'];
```
Huachao, line 55: For this case, do we need to do the URL encode work explictly?

bolthorn91, line 55: What about putting all arguments in array in order to scale or abstract if we need it? for example:

bolthorn91:
```suggestion
        const dataKeys = ['d', 'data', 'data-ascii', 'data-binary', 'data-raw', 'data-urlencode'];
        let body = parsedArguments[dataKeys.find(key => !!parsedArguments[key])];
```

bolthorn91 (commented):
Just a little improvement for future purposes

src/utils/curlRequestParser.ts
mikecanann: yeah, that has been added to the PR

bolthorn91 (approved):
GJ!

Issue #766, opened 2021-02-05
PROPFIND (webdav/caldav)


mitsol:
Would there be the possibility to also extend the HTTP operations with the ones from webdav/caldav (PROPFIND, PROPPATCH, REPORT, MKCOL, MKCALENDAR, ACL). I would find it a nice feature for testing caldav applications.

georghub:
Same for me, additional http methods would be very helpful for testing WebDAV calls:
PROPFIND, PROPPATCH, MKCOL, LOCK, UNLOCK, SEARCH

PR #1030, opened 2022-07-11
Add support for WEBDAV verbs

ternag:
This is a suggestion to address the request ( #766 ) for support of the http verbs used by webdav (and caldav).


ternag:
Not that I know of. I have used it for testing webdav enpoints without any problems.

Huachao:
@ternag Merged, thanks

PR #1107, opened 2022-12-14
Use token response expires_in to remove token from cache

laurynasr:
$aadV2Token helper is using the same token even after it expires. This change removes the token from cache after the `expires_in` time expires.


Huachao (approved):

Huachao:
LGTM

Huachao:
@laurynasr thanks for your contribution

PR #1117, opened 2023-01-11
Support host aliases in client certificate configuration

mpcjanssen:
We are using an enterprise service bus where different users have different authorisations and use different client certificates. So I  need to be able to configure multiple client certificates for the same host.  

This PR allows changing the host of the http request in the certificate settings so one can define an alias for every different certificate on the same host. 

Another option to achieve the same effect would be to be able to define a certificate on every request but that is a bigger change and requires more duplication in the `http` files themselves.

Example use:

`config.json`

``` json
{
    "rest-client.certificates": {
        "esbnode_certa": {
            "hostname": "esbnode",
            "pfx": "user_a.p12",
            "passphrase": "pw_a"
        },
        "esbnode_certb": {
            "hostname": "esbnode",
            "pfx": "user_b.p12",
            "passphrase": "pw_b"
        }
    }
}
```

And then in the `http` file:

``` http

### with cert a
GET https://esbnode_certa/operation

### with cert b
GET https://esbnode_certb/operation


```


nt126716:
Hi @mpcjanssen 
Instead of having hostname in certificate block, can we have requestUrl as below

```json
{
    "rest-client.certificates": {
        "certa": {
            "requestUrl": "https://example.com/path1",
            "pfx": "user_a.p12",
            "passphrase": "pw_a"
        },
        "certb": {
            "requestUrl": "https://example.com/path2",
            "pfx": "user_b.p12",
            "passphrase": "pw_b"
        }
    }
}
```

And then users can simple use regular syntax or can follow the same syntax of having per-request settings standard as
Regular Syntax:
```http
GET https://example.com/path1/operation

###
GET https://example.com/path2/operation
```

Per-request settings:
```http
# @cert certa
GET https://example.com/path1/operation

###
# @cert certb
GET https://example.com/path2/operation
GET
```

PR #1137, opened 2023-03-13
implement loading indicator

### LINKED ISSUE #848: Feature-Request: Display Request-Status when action is performed (loading icon or something)
Thank you for providing such a lovely tool, i am currently using it a lot and trying to move from postman.

I think this feature isn't present yet within the plugin, but i am using the following setup:
- REST Client Version: 0.24.5
- VSCode Version: 1.56.2
- OS Version: Windows 10

### Current state
When performing a request (which can take a long time), i currently cannot see what the current request-status is or if my action has been performed at all (or if my pc is stuck/frozen 😄).

### Desired state
When performing a request, the user should be able to see if he started/performed this action by displaying a loading-icon (optional: with according loading-text message)

I see 3 possibilities for the implementation:
1. add loading icon directly after the "Send Request"-Button (see screenshot at bottom)
2. Open the Response-window before the actual response comes in and display some kind of loading-screen
3. A notification window, which pops up (at bottom right) and displays the current status (like when the vscode-update is performed)

I don't know what the best solution would be to tackle this problem, but any described solution (or combination of these) would already bring benefits.

### Description
When performing a request, which takes a long time (like 20+ seconds), i cannot confirm if my performed actions have been correct or my pc has some kind of issue which causes the request to not be performed at all. Most of the time, i try to click 2 to 5 times on the same "Send Request"-Button, which doesn't help with the performance. 😅 
In this case, my API-Calls were POST-Requests which were performing various image processing tasks - this could take up to a minute.

To solve this problem, the current state of the request should be visible by using a loading-icon or some kind of additional message/notification.

I think this feature should be relatively easy to implement once the best solution has been decided and contributes a lot to the user experience of this tool.

image

I would love to get some feedback or other suggestions on this - thank you very much 😄 
BR, Edvin

quanterion:
Display information about currently in-progress request

image

fix https://github.com/Huachao/vscode-restclient/issues/848

### DIFF
--- a/src/controllers/requestController.ts
+++ b/src/controllers/requestController.ts
@@ -2,6 +2,7 @@ import { ExtensionContext, Range, TextDocument, ViewColumn, window } from 'vscod
 import Logger from '../logger';
 import { IRestClientSettings, RequestSettings, RestClientSettings } from '../models/configurationSettings';
 import { HistoricalHttpRequest, HttpRequest } from '../models/httpRequest';
+import { HttpResponse } from '../models/httpResponse';
 import { RequestMetadata } from '../models/requestMetadata';
 import { RequestParserFactory } from '../models/requestParserFactory';
 import { trace } from "../utils/decorator";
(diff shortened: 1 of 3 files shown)


Puliczek:
+1

uudashr:
+1
I would love this to be implemented as well

Issue #740, opened 2020-12-29
Allow for a dotenv per environment


elpablete:
Many times I find myself with an .env file per environment (local.env, test.env, prod.env), that way I can use the same names on my code and still not commit any values to repos. Right now there is no way to do so, but a more powerful feature would be to allow a different .env file to be defined in each environment (even in $shared)

philwrenn:
Something like this would be helpful for me too. I'm not currently using different .env files, but I would like to store a few .http files in a folder. That would mean they wouldn't be in the project root with my .env file.

If we could provide a relative path to a specific .env file, then @elpablete and my issue would possibly both be addressed.

FDiskas:
usually such files are named like so 
```
.env
.env.local
.env.production
```

wekempf:
I'd love this as well, but until (if?) this is done, there's a workaround. In your `settings.json` you define your environments like this.

```
"dev":
{
    "secret": "devSecret"
},
"prod":
{
    "secret": "prodSecret"
}
```

In your `.env` file you'd then have this.

```
devSecret=xyzzy
prodSecret=plough
```

Finally in your http file you'd have this.

```
@secret = {{$dotenv %secret}}
```

A little more complicated to setup, but this allows the values you use to change when you change the environment. This is untested and may need some tweaking, but the idea should work.

monisnap-paul:
Would it be possible to try to load first .env.<environmentName> with environmentName the name of the currentEnvironment  
(e.g. .env.local, env.dev, ...) and if this file does exist, load the .env file ?

Nilegfx:
I would love to have this feature too!
Maintainers, could you please review and merge the PR mentioned here?

PR #1149, opened 2023-05-01
[Feature] add support for vscode env in dotenv

nicksmider:
This PR allows the users to create .env files for each environment that exists within their vscode settings. 

Screenshot 2023-05-01 at 2 51 24 PM

Behavior: When walking back up the path, when either `.env` or `.env.{vscode-env}` is found, stop and prefer `.env.{vscode-env}`.

Use case:
I want to specify an auth token in gitignored env files because the repo's settings.json is checked into source control already so we cannot place them in there safely.

Closes https://github.com/Huachao/vscode-restclient/issues/740

### DIFF
--- a/src/utils/httpVariableProviders/systemVariableProvider.ts
+++ b/src/utils/httpVariableProviders/systemVariableProvider.ts
@@ -6,6 +6,7 @@ import * as fs from 'fs-extra';
 import * as path from 'path';
 import { Clipboard, commands, env, QuickPickItem, QuickPickOptions, TextDocument, Uri, window } from 'vscode';
 import * as Constants from '../../common/constants';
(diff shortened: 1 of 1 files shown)


shaunakv1:
Will this read the .env file in the root of code that's currently open?

Huachao (approved):

Huachao:
@nicksmider merged, thanks

Huachao:
@FDiskas the `dotenv` is already used by the project

sayedihashimi:
Based on the use case outlined by @nicksmider I think a better approach would be to first load the .env file and then load values from .env.local as overrides. From the description it sounds like users will need to copy all values from the .env file and now there are two things to maintain.

PR #1197, opened 2023-09-25
[Feature] Read aad2 scopes from environment variables

NickRimmer:
Hi @Huachao,
Thank you a lot for your work and such a great plugin. Looks very promising for me to use. 

I would like to suggest a very tiny PR with a feature. I found that all AAD settings can be set per environment except scopes. But on AAD, multiple app registrations can have different scopes (e.g. Dev/User.Default, Prod/User.Default, etc.). 

It would be awesome if you could approve my PR and add this feature to your extension. I wasn't able to build `vsix` package from your current master branch, so based it on a commit with tag `v0.25.0` and tested it on my local.

Thank you (;


Huachao:
@NickRimmer Thanks, merged

PR #1200, opened 2023-09-27
Adds ability to force json or xml parsing during templating

gislikonrad:
When requesting HTTP endpoints not under your control, you could come across an endpoint that doesn't respond with the correct content-type header value. For example getting an xml body, but the header says it's text/plain. You could also get a json body and there is no content-type header. 

With this PR, I add a way to force parsing as json or xml when templating.

For JSON it would be like this: `{{previous_request_name.response.body.asJson.$.data.id}}`
For XML it would be like this: `{{previous_request_name.response.body.asXml.//data/@id}}`


JerryJian:
I need this feature, because my response header is text/html, but the content is json format. Looking forward to the merger this PR, thanks. @Huachao

Huachao (approved):

Huachao:
@gislikonrad @JerryJian  merged, thanks

PR #1278, opened 2024-06-21
Preserve multiple `Set-Cookie` headers

### LINKED ISSUE #722: Set-Cookie response headers are grouped
- REST Client Version: 0.24.4
- VSCode Version: 1.51.1
- OS Version: windows 10

Steps to Reproduce:

1. imagine this simple express script...

```
var express = require('express');
const app = express();
app.get("/", (req, res) => {
  res.cookie('n1', 'v1');
  res.cookie('n2', 'v1');
  res.sendStatus(200);
});
app.listen(8080, () => {
  console.log(`listening on port 8080`);
});
```

2. simple GET request will group `Set-Cookies` headers into one as follows;

```
HTTP/1.1 200 OK
X-Powered-By: Express
Set-Cookie: n1=v1; Path=/,n2=v1; Path=/
Content-Type: text/plain; charset=utf-8
Content-Length: 2
ETag: W/"2-nOO9QiTIwXgNtWtBJezz8kv3SLc"
Date: Mon, 30 Nov 2020 12:08:27 GMT
Connection: close

OK
```

NOTE almost same request from curl...

```
λ curl -v http://localhost:8080
* Rebuilt URL to: http://localhost:8080/
*   Trying ::1...
* TCP_NODELAY set
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET / HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.55.1
> Accept: */*
>
< HTTP/1.1 200 OK
< X-Powered-By: Express
< Set-Cookie: n1=v1; Path=/
< Set-Cookie: n2=v1; Path=/
< Content-Type: text/plain; charset=utf-8
< Content-Length: 2
< ETag: W/"2-nOO9QiTIwXgNtWtBJezz8kv3SLc"
< Date: Mon, 30 Nov 2020 11:59:57 GMT
< Connection: keep-alive
<
OK* Connection #0 to host localhost left intact
```

jordanbtucker:
This PR ungroups `Set-Cookie` headers since each header should appear independently. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie):

> To send multiple cookies, multiple `Set-Cookie` headers should be sent in the same response.

For example, the following is correct:

```
Set-Cookie: a=b
Set-Cookie: c=d
```

While the following is incorrect:

```
Set-Cookie: a=b,c=d
```

This PR achieves this by:
- Adding an exported `formatHeaders` function in `utils/misc.ts` that treats `set-cookie` as a special case.
- Removing the `private formatHeaders` function from `HttpResponseTextDocumentView` and updating all calls to use the new utility `formatHeaders` function.
- Removing the `private static formatHeaders` function from `HttpResponseWebview` and updating all calls to use the new utility `formatHeaders` function.
- Updating `getFullResponseString` in `HttpResponseWebview` to use the new utility `formatHeaders` function.
- Updating `createRequestInTempFile` in `HistoryController` to  use the new utility `formatHeaders` function.

Fixes #722

### DIFF
--- a/src/controllers/historyController.ts
+++ b/src/controllers/historyController.ts
@@ -6,6 +6,7 @@ import * as path from 'path';
 import { QuickPickItem, window, workspace } from 'vscode';
 import { HistoricalHttpRequest } from '../models/httpRequest';
 import { trace } from "../utils/decorator";
+import { formatHeaders } from '../utils/misc';
 import { UserDataManager } from '../utils/userDataManager';
 
 dayjs.extend(relativeTime);
@@ -62,14 +63,7 @@ export class HistoryController {
     private async createRequestInTempFile(request: HistoricalHttpRequest): Promise<string> {
         const file = await this.createTempFile();
         let output = `${request.method.toUpperCase()} ${request.url}${EOL}`;
-        if (request.headers) {
-            for (const header in request.headers) {
-                if (request.headers.hasOwnProperty(header)) {
-                    const value = request.headers[header];
-                    output += `${header}: ${value}${EOL}`;
-                }
-            }
-        }
+        output += formatHeaders(request.headers);
         if (request.body) {
             output += `${EOL}${request.body}`;
         }
--- a/src/utils/misc.ts
+++ b/src/utils/misc.ts
@@ -33,6 +33,32 @@ export function removeHeader(headers: RequestHeaders | ResponseHeaders, name: st
     }
 }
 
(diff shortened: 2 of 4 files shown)


Huachao:
@jordanbtucker Thanks, merged

PR #1302, opened 2024-09-20
Remove incorrect question mark from requestline regex

### LINKED ISSUE #955: API response not highlighted
- REST Client Version: v0.24.6
- VSCode Version:Version: 1.63.2 (system setup)
- OS Version: Windows_NT x64 10.0.19044 (Win 10 x64 21H2)

# Steps to Reproduce
1. Make a POST request like below
```
POST https://huoche.tuniu.com/yii.php?r=train/trainTicket/getTickets&primary[departureDate]=2022-01-01&primary[departureCityName]=杭州&primary[arrivalCityName]=上海虹桥
```
The response body has no highlighting.

2. Make a GET request like below
```
GET http://ip.jsontest.com/
```
The response is probably highlighted.

Both requests seem to be smaller than `rest-client.largeResponseBodySizeLimitInMB`.
# Desired Output
Both requests are properly highlighted.

danlapid:
Fixes #955

The current regex matches 
`HTTP/1.1 200 OK` as `requestline` instead of`responseline` this is because the regex is too lenient allowing no METHOD field.
According to the RFC (https://datatracker.ietf.org/doc/html/rfc2616#section-5.1) METHOD has to exist and be not empty for the request line.
You can test the regex at https://regex101.com/

### DIFF
--- a/syntaxes/http.tmLanguage.json
+++ b/syntaxes/http.tmLanguage.json
@@ -258,7 +258,7 @@
           ]
         }
       },
-      "match": "(?i)^(?:(get|post|put|delete|patch|head|options|connect|trace|lock|unlock|propfind|proppatch|copy|move|mkcol|mkcalendar|acl|search)\\s+)?\\s*(.+?)(?:\\s+(HTTP\\/\\S+))?$",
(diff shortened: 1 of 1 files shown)


danlapid:
@Huachao, would appreciate a review, thanks 😄

Huachao (approved):

PR #1313, opened 2024-10-18
Add "Copy Exchange" command

michael-wolfenden:
Allows copying the full request and response to the clipboard


domoran:
I like this  feature. Could you add a "Copy Response Body" command too? This is especially good for copying JSON results without having to switch the preview to "body only" ...

